PR #3, opened 2021-09-17
Add Features.md for a comprehensive list of features

dri94:
It would be nice to have a comprehensive list of features so users can make a more informed decision about which fork they wish to use and why. These are the features I know about, if there are more, let me know


sunnyhaibin (approved):
Approved to merge. Thanks!

PR #4, opened 2021-09-19
Sidebar temperature gauge fix

JamesKGithub:
Fixes the boot error with the Totota MADS addition and increases the temperature gauge font size.


sunnyhaibin (changes requested):
Please remove `panda/board/safety/safety_toyota.h` from this PR and we should be able to merge!

PR #5, opened 2021-09-20
Temp gage fix

JamesKGithub:
Fixes the temperature gage to be more in size of the other boxes.


sunnyhaibin (approved):
Approved for merging. Thanks!

PR #8, opened 2021-09-21
dynamic laneless button 3/3

twilsonco:
three-part PR to reposition the laneless button and make it dynamic so it works on C3 with nav mode. 

The button press only works when you press the laneless button _in nave mode_, not otherwise, but that's because all screen presses in normal (sidebar hidden) mode are hijacked to switch to nav mode.

Looks pretty good.

I also blew it up a bit.

IMG_0781
IMG_0778
IMG_0785


sunnyhaibin:
Looks great! Thanks for contributing :)

PR #9, opened 2021-10-06
Suppress C3 Nav toggle when DLP button is pressed

twilsonco:
On C3, any mouse press causes the onroad gui to 3-way toggle between map, sidebar, neither, and the laneless button only responds when pressed in nav mode.
Nav is now only toggled when the mouse press is not on the laneless button.

@twilsonco
  
Author
twilsonco commented now
Also with this the button button works when in nav or normal (no sidebar) mode.


sunnyhaibin:
Thanks for catching this!

PR #13, opened 2021-11-15
Cleaner UI

FrogAi:
Makes the settings button look less cluttered especially with the new "Features" panel and upcoming "Navigation" panel.


sunnyhaibin:
Hmm not sure why this PR showed that I closed it. I will add this in the next release and mention this PR for sure! Thanks for the PR! :)

PR #23, opened 2022-01-01
Update Subaru car list in Force Car Recognition

FrogAi:
Should help clear up confusion for users on why their car isn't listed.


sunnyhaibin:
Thanks! :)

PR #30, opened 2022-01-13
Higher torque limit (3071) for 2018 Crosstrek / 2017 Impreza @ClockeNessMnstr

Moodkiller:
Discussion / original PR: https://github.com/martinl/openpilot/pull/71

Supported / tested EPS':

- b'z\xc0\x04\x00',
- b'z\xc0\x08\x00',
- b'z\xc0\x00\x00',

Personally this works well for my 2017 Subaru Impreza and it now makes turns which used to exceed the comma default limits.

All credit goes to @ClockeNessMnstr for this.


ClockeNessMnstr:
Crosspreza, I like it lol

Discussion is on Martin's PR, he's cautious of panda changes. If we can get some logs of the max torque from someone's cross track or Impreza with a different EPS from these we can confirm that the torque is actually the same just with different values. 

One last note is the fingerprinting. I saw it was getting multiple fingerprints in the unit test. I believe this works fine and that it's just the fuzzy fingerprinting that's failing the unit test, but I want to figure out how to enforce the EPS match for these fingerprints.

Moodkiller:
Haha, ty ty. Feel free to steal for whatever you might need it for. 

One other quirk I've found and confirmed with prlifestyle is that, with sunnys release anyways, we need to set the car to `SUBARU CROSSTREK 2018` in order for the higher torque limits to apply. If we set as an Impreza variant, it seems to revert to the original value. Nor sure if fuzzy fingerprinting you mentioned but yeah.

ClockeNessMnstr:
Sort of, setting to "Crosstrek" is how it works. Treats the sets of eps as different vehicles. New set gets higher torque. 

Fuzzy fingerprinting matches both without an eps because I just copied the other modules from Impreza.

But the EPS needs to be known so I think I can find a way to restrict that with storming upstreamable

sunnyhaibin:
Will merge when commaai/openpilot#23530 is merged with stock openpilot.

Moodkiller:
Fair enough. I see there has been quite a few changes. I will update this with those changes so that I can test it for myself as well.

Moodkiller:
Good news, latest commit (https://github.com/sunnyhaibin/openpilot/pull/30/commits/d5c716bb0a50a0e5fb816ba84c08ee6445fd85b7) works. This is now the same code used for the PR on comma's git. 

Only thing is find some tuning that works for sharp corners going +-45KM/h as it doesnt seem to damper nicely. I.e car nears the centre line during the turn > touches > increases turning moving away from center line > relaxes > nears center line again > repeat.

ClockeNessMnstr:
> Good news, latest commit ([d5c716b](https://github.com/sunnyhaibin/openpilot/commit/d5c716bb0a50a0e5fb816ba84c08ee6445fd85b7)) works. This is now the same code used for the PR on comma's git.
> 
> Only thing is find some tuning that works for sharp corners going +-45KM/h as it doesnt seem to damper nicely. I.e car nears the centre line during the turn > touches > increases turning moving away from center line > relaxes > nears center line again > repeat.

You can cut the old tune by 1/3 to be the same as the old stock tune. 
I don't get much time to drive, but once I get a chance to do some testing for the upstream PR I'll take a crack at the tune as well. 

```
ret.lateralTuning.pid.kf = 0.0000333
  ret.lateralTuning.pid.kpV, ret.lateralTuning.pid.kiV = [[0.133, 0.2], [0.013, 0.02]]
```

Moodkiller:
Ah perfect, thank you for that! I tried a few of your tune commits (on the `comma3` branch) but didn't find one with the same "confidence" as stock tune + more torque. 

I'll plug in those values and give it ago tomorrow - fwiw I drive to and from work during the weekdays so have plenty of opportunities to test if you want to try something and have written feedback would good enough to work from.

Moodkiller:
> ```
> ret.lateralTuning.pid.kf = 0.0000333
>   ret.lateralTuning.pid.kpV, ret.lateralTuning.pid.kiV = [[0.133, 0.2], [0.013, 0.02]]
> ```

This was a good adjustment in terms of the dampening in corners. Only feedback now would be if it could turn sooner / hold a line tighter into the corner. Looking at comma's wiki for tuning, the crude method anyways, increasing `steerActuatorDelay` should make it turn sooner but I found this to not be the case. 

I see you have some very different values on your C3 branch in comparison. Are you find the same issues as I or are you testing something else?

ClockeNessMnstr:
A little bit of both. I believe the stock controller doesn't allow for high enough integrator values. I don't have the dedicated time to tune right now so I'm slowly testing different values. 

 increased feed forward may help push it through turns though. In fact you can turn p and I to zero and tune feedforward. It won't really be able to take the turns With the right timing  but you should get roughly the torque you want

PR #31, opened 2022-01-13
Feature: `CameraOffset`

Moodkiller:
Couple of  notes this with this PR: 

1) I set a few params to be [ON] by default in between installs, namely:
- Training is marked as completed so don't have redo between installs.
- Terms and Conditions have been accepted.
- Turn on:
   - Lane Departure Warning, 
   - Right hand drive 
   - Metric units.
- Enable `Quiet Drive` by default.
- Enable `SSH` by default.

2) This also implements PR: https://github.com/commaai/openpilot/commit/233a6e4a625a8ef145904522e4663177bcd3d65d

3) Visual change: Increased the font size of the +/- buttons slightly.

4) Line 55-64 in `selfdrive/controls/lib/lane_planner.py` is from DP's Camera Offset feature. I think this can be altered to update the variable without turning off and on the car (TODO) 

5) This is compatible with all hardware.

### DIFF
--- a/CHANGELOGS.md
+++ b/CHANGELOGS.md
@@ -31,6 +31,7 @@ sunnypilot - Version 0.8.12-3
 
 sunnypilot - Version 0.8.12-2
 ========================
+* IMPROVED: Update Subaru car list in Force Car Recognition (#23) thanks to FrogAi!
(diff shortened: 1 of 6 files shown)


sunnyhaibin:
Thanks! :)

PR #45, opened 2022-02-23
Shane's ZSS startup alert

davidwuya:
https://github.com/sshane/openpilot/pull/430


davidwuya:
yeah this is not working right now - ZSS works fine on my TSS-P '16 Prius, just the alert is not working

PR #55, opened 2022-04-15
Dismiss PrimeAd & add third block (@sshane)

Moodkiller:
Modified code from Shanes original commit to work with this release: https://github.com/sshane/openpilot/pull/439/files

Shows how to get prime, comma points and third block that can be used for whatever (shutdown and restart buttons here would be nice).

Don't have prime so cant test whether this behaves correctly with a subscription but it _should_.

User will be shown the full PrimeAd once and then not again until re-install. Will always show how to get prime / subscription status. Could initialize branch with `PrimeAdDismissed` set  to `1` to avoid having to hit the `Dismiss` button once.


Moodkiller:
> Too much has changed since the last time the branch was updated. I'll include it in the upcoming sunnypilot v0.8.14. Thanks for the PR :)

No problem, sounds good! Looking forward to it.

PR #59, opened 2022-04-30
Add additional 2022 Santa Fe FW

chrsfr:


sunnyhaibin:
You could use Force Car Recognition to force recognize your vehicle.

The FW version should be PR'ed to upstream.

Thanks for the PR!

PR #66, opened 2022-06-08
AtL's INDI Corolla TSS2 Tune v2.0

AtLmd:
A vastly improved tune for corolla TSS2 than PID

### DIFF
--- a/selfdrive/car/toyota/interface.py
+++ b/selfdrive/car/toyota/interface.py
@@ -166,7 +166,8 @@ class CarInterface(CarInterfaceBase):
       ret.steerRatio = 13.9
       tire_stiffness_factor = 0.444  # not optimized yet
       ret.mass = 3060. * CV.LB_TO_KG + STD_CARGO_KG
-      set_lat_tune(ret.lateralTuning, LatTunes.PID_D)
+      set_lat_tune(ret.lateralTuning, LatTunes.INDI_COROLLA_TSS2)
+      ret.steerActuatorDelay = 0.41 # AtL's INDI Corolla TSS2 Tune v2.0
       ret.wheelSpeedFactor = 1.035
 
     elif candidate in [CAR.LEXUS_ES_TSS2, CAR.LEXUS_ESH_TSS2]:
--- a/selfdrive/car/toyota/tunes.py
+++ b/selfdrive/car/toyota/tunes.py
@@ -24,7 +24,7 @@ class LatTunes(Enum):
   PID_L = 13
   PID_M = 14
   PID_N = 15
-
+  INDI_COROLLA_TSS2 = 16
 
 ###### LONG ######
 def set_long_tune(tune, name):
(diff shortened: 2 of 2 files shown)


AtLmd:
Do you have the link to that old tune?

bscholer:
After running this tune for a bit, I'm confident in saying that it is a safety hazard and should not be used. I encountered multiple lane ejections, and I believe that any officer in their right mind would have written me a DUI for the severe oscillations. 

I'm of the opinion that birdman's tune, despite being old, is still the best for straight roads and moderate curves. While you admittedly lose a fair amount of torque, I personally prefer that if it means a smoother, more comfortable ride 99% of the time. My region (northern/central Indiana) is very flat with lots of straight roads, and the smoothness of birdman's tune is fantastic. That said, lowering `actuatorEffectiveness` on birdman's a bit helps with oscillations even more. Will make a PR for this soon.

AtLmd:
I’ve got a touring sports and my tune for it is way better than birdman. Maybe the touring sports tune is different from the hatchback or sedan version?

AtLmd:
At what speed does your car oscillate severely?

bscholer:
Interesting--I'm in a 22 Corolla Hatch XSE. Oscillations occur at all speeds, but it gets worse as I increase the speed. I didn't go above 60mph with it, but it would start fine, and then start oscillating back and forth at an increasing rate until eventually either leaving the lane or getting very close to it, forcing me to take control. 

Looks like the Touring Sports has a curb weight of `2821 lbs`, and `interface.py` shows `3060 lbs`. I'm not sure if this plays a role in oscillations though.

I do think that `steerActuatorDelay` may be to blame though. Your tune uses `0.41`, whereas birdman's uses `0.22`. I have not tried your tune with an adjusted `steerActuatorDelay`, but I will give this a shot if I have time.

Have you tried adding breakpoints to your tune instead of running straight gains?

In any case, if it works for you, I'm glad. I would just be hesitant to merge if it were my fork. 🙂

AtLmd:
I did not have to try any breakpoints as it was really smooth from the start for all the speed range, no oscillations. I still don’t know why does it scale linearly but It works wonders for me.

AtLmd:
You said "I'm confident in saying that it is a safety hazard and should not be used." but you did tamper with the tune.
I would be interested in your feedback retrying it for what it is.

bscholer:
@AtLmd I have no idea what I screwed up when I tested this first, but I gave it another go, and it's _great_. I owe you an apology, I didn't notice any of the crazy oscillation I experienced last time. I have a pretty tricky test route I have it drive, and it did pretty well.  

I drove the route once with `steerActuatorDelay` set to `0.41`, and once with it set to `0.22`. I think that somewhere between the two is probably the right number, at least for my car and preference. Probably around `0.35-0.38`. Will post these routes soon.

Two prevailing issues though:
1. The steering is pretty jumpy. I'm not sure that there is a way to smooth it out though without sacrificing torque.
2. It struggles a bit in very sharp turns, although this got noticeably worse when I set `steerActuatorDelay` lower.

I will be using this tune going forward though, you did a nice job with it! Thanks!

AtLmd:
Glad it worked. You have to take into account that the tune was made for OP 0813 that’s why I opened a PR only on sunnyhaibin 0812 (as he ported the 0813 model on his fork).
Also to notice : you have to differenciate problematic behaviours coming from the tuning or the model itself. My results on OP0813 lanefull are better than laneless (oscillates on laneless so it’s a model "characteristic")
Would you mind trying OP 0813 **lanefull** on your C3 to test it? I believe it will fix the behaviours you experienced.

AtLmd:
`steerActuatorDelay` emits controls ahead of plan, you know it’s right when it enters a turn at the right time. It has nothing to do with torque.
- If turning too early, decrease steerActuatorDelay
- If turning too late, increase steerActuatorDelay

If you feel like you car doesn’t turn well you can set a fix steeratio by bypassing the learner.
I set mine at `ret.steerRatio = 15`
https://github.com/eyezenheim/eyezenpilot/commit/02f18932135ae7eabe21dcad9d95ef62b50d5a15

AtLmd:
`steerRatio` has nothing to do with you car mileage. Steer ratio learner can mess up your drive as it learns it through every drive. If you don’t fix it there is no point in trying to tune as you don’t separate the variables and you can’t isolate the problem.

Stupefacient:
Try a delay of 0.33

bscholer:
Never got around to trying 0.8.12, but using .33 seems to work really well! Tested it for ~300 miles on lots of different roads.

sunnyhaibin:
Thanks for the PR. INDI lateral control has been deprecated and is superseded by Torque Lateral Control in the latest sunnypilot/openpilot.

PR #70, opened 2022-07-01
Fix Honda not braking down to 0

steveham3:
Fixes openpilot not braking to zero if using Honda's stock ACC and the speed goes below the ACC minimum engage speed

### DIFF
--- a/selfdrive/controls/controlsd.py
+++ b/selfdrive/controls/controlsd.py
@@ -579,7 +579,7 @@ class Controls:
 
     if not self.joystick_mode:
       # accel PID loop
-      long_active = self.active and CS.cruiseState.enabled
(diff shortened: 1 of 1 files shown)


luisgrases:
Do you know if this or something similar can be done for Toyotas that disengage acc at speeds below 30kph?

steveham3:
This fix is in a file that isn’t Honda specific. It may work for Toyota too. If it doesn’t then I would contact spektor56 as I worked with him to develop this fix.

On Jul 15, 2022, at 1:20 AM, Luis Grases ***@***.***> wrote:

﻿

Do you know if this or something similar can be done for Toyotas that disengage acc at speeds below 30kph?

sunnyhaibin:
Thanks for the PR. This behavior should be fixed in the latest sunnypilot.

PR #72, opened 2022-07-03
Added support for Tucson 2019 Diesel

fsalsubaei:


sunnyhaibin:
This vehicle is now supported in upstream. The latest [`test-c3`](https://github.com/sunnyhaibin/sunnypilot/tree/test-c3) or [`test-c3-20221205`](https://github.com/sunnyhaibin/sunnypilot/tree/test-c3-20221205) should have support for this vehicle as well.

Thanks for the PR!

PR #73, opened 2022-07-07
Readme.md Typo Fixes and Suggestions

akhairaddin:
This is just a suggestion for fixes to some typos
<> = Means Delete
()    = Means Add
---------------------------------------------------------------------------------------------------------------------------------------------
Line 74: **Driving Screen Off Timer** - Turn(s) off the device screen or reduce(s) brightness to protect the screen after car starts

Solution: add an "s" to "Turn" and "reduce"

Line 94: 2. `OFF`: `BRAKE pedal` press will NOT (pause) sunnypilot Automatic Lane Centering; `BRAKE pedal` release will NOT resume ACC/SCC without an explicit entry **(strictly enforced in panda safety code)**

Solution: Add "pause" between "NOT" and "sunnypilot"

Line 100: Dynamic Lane Profile (DLP) aims to provide the best driving experience <with> (at) staying within <the> a lane confidently.
Dynamic Lane Profile allows sunnypilot to dynamically switch between lane profiles base(d) on lane recognition confidence level on road.

Solution: Replace "with" with "at" as well as replace "the"  with "a" and also add a "d" in front of "base"

Line 107 - 113: "Disable use of lanelines" toggle is not available on 0.8.14 and above

Lines 132,135,137: **Note: Require(s) data connection**

Solution: Add "s" infront of "Require"

Line 137: Hands on Wheel Monitoring - <Monitor and alert when driver is not keeping the hands on the steering wheel>

Solution: Reword this sentence to "Monitors and alerts the driver when their hands have not been on the steering wheel for an extended time"

Line 144: When using Speed Limit <Control>, Vision(,) or Map based Turn control, you will be setting the "MAX" ACC speed on the sunnypilot display instead of the one in the dashboard. The car will then set the ACC setting in the dashboard to the targeted speed, but (will) never exceed<ing> the max speed set on the sunnypilot display. A quick press of the RES+ or SET- buttons will change this speed by 5 MPH or KM/H on the sunnypilot display, while a long deliberate press (about a 1/2 second press) changes it by 1 MPH or KM/H. DO NOT hold the RES+ or SET- buttons for longer that a 1 second. Either make quick or long deliberate presses only.

Solution: Delete the unnecessary "Control", Add a comma after "Vision" as well as add "will" after "but" and delete "ing" from "exceed"


sunnyhaibin:
This PR is being submitted from the wrong branch, so I can't merge this.

I'll make the suggested changes in the main branch. Thank you!

PR #74, opened 2022-07-12
Persistent tethering state when reboot

rogerioaguas:
In this feature i was _added the Hotspot state when boot_, in last version u did always turn on hotspot when _Comma reboot_, now the comma boot with last state of tethering. 

The code try bring the very simple solution to this feature.

Thx


sunnyhaibin:
Thanks for the PR!

0.8.14 branches are no longer maintained. I'll have this feature implemented in the latest `test-c3` branch with credits.

PR #85, opened 2022-10-27
WIP - HKG CAN-FD - initial dash speed in SP

ioniq5-cz:
@sunnyhaibin this is WIP and I open the PR  for you to review my code changes only at this stage.


sunnyhaibin:
This method won't work. cruiseState.speed still needs to be kept the same.

Do you have a route with this DBC change, and set speed on the dashboard set to above 127 km/h?

ioniq5-cz:
Route is here in last comment https://github.com/commaai/opendbc/pull/729#discussion_r1007825878

ioniq5-cz:
Also @sunnyhaibin - have a look at https://github.com/ioniq5-cz/sunnypilot/blob/test-c3/selfdrive/car/hyundai/carstate.py - I think I am super close now ;) this is still WIP 🚧

PR #88, opened 2022-12-01
Battery Protection

EdwardApollo:
Battery Voltage Floor to 11.9, LPF Gain = 0.011 (Tau=45s)

This significantly improves logic to shutdown the device to prevent excessive battery degradation.

Min battery voltage is a great candidate for an additional toggle/input


EdwardApollo:
Ref: https://github.com/commaai/openpilot/pull/26640 for in depth discussion

PR #91, opened 2022-12-04
Add 2020 Lexus UX 200 (non-hybrid) support

kikiyinfu:
Added 2020 Lexus UX 200 (non-hybrid) fingerprinting


sunnyhaibin:
Thanks for the PR. The FW versions have been added in the latest sunnypilot/openpilot (https://github.com/commaai/openpilot/pull/26669), therefore this PR is now redundant.

PR #94, opened 2022-12-15
Integra 6MT 2023

tapperoo42:
Grabbed the fingerprint from my Integra. I wasn't sure if comma would accept it since it's a MT.

Fingerprint is from route 1b7349c28512586d|2022-12-14--09-34-05

Made a dedicated dbc file because of the missing 0x191 and 0x158 ("GEARBOX" and "ENGINE_DATA" for Automatics / CVT).

LKAS stops @ 40 mph; AAC @ 25 mph (LKAS might be able to work down to 25?)  but given the high start speed, maybe "ENGINE_DATA" "XMISSION_SPEED" might not be necessary.


sunnyhaibin:
Hey, can you please re-open this PR against the `master` branch? Whenever this branch gets updates it will auto-close PRs. Thanks!

tapperoo42:
Sure thing - There are definitely issues with the PR from when I tried to get things working. The car doesn't have torque values and several other items that still need to be defined that I didn't know to look into. I thought it would just steal them from the 2022 civic.

PR #103, opened 2023-02-03
Added iq5 fwd camera fingerprint

Cx01N:


sunnyhaibin:
Thanks for the PR. The FW version in this PR has been merged in the latest sunnypilot/openpilot (https://github.com/commaai/openpilot/pull/27137), therefore it is redundant.

PR #114, opened 2023-02-17
Hyundai: Enhanced SCC (ESCC) Radar Interceptor support

sunnyhaibin:
Detect Enhanced SCC (ESCC) Radar Interceptor on `0x2AB`.

ESCC is a hardware modification that a White Panda with custom firmware injected between the C-CAN H and L between the car's SCC radar and the car. The hardware modification is independent and does not require major refactor to openpilot's control and panda safety code. This must be built/purchased and installed separately.

When openpilot longitudinal is enabled, ESCC firmware switches automatically to block all SCC-related messages (`SCC11, SCC12, SCC13, SCC14`) to allow openpilot send its own longitudinal commands to the car. This keeps the car's SCC radar enabled and allows forwarding of AEB/FCW events to the car.

**Prerequisite of**
- https://github.com/sunnyhaibin/panda/pull/2


sunnyhaibin:
Drove for 20 minutes, was able to generate a FCW event from the radar with openpilot longitudinal enabled.

PR #116, opened 2023-02-18
Change osm route nodes distance algorithm

pfeiferj:


pfeiferj:
Marking as draft as I'm still testing and need to prove objectively that this corrects issues regarding route selection. Created the pr against this branch for now since it's the one being used for testing, but I can create a pr against master once ready to merge.

PR #125, opened 2023-03-10
Update hyundai_enable_radar_points.py

WizardSource:
2020 Sonata FCUP support


sunnyhaibin:
Thanks for the PR. Do you have a route with raw logs uploaded that has radar tracks enabled?

sunnyhaibin:
This was added in the latest sunnypilot/openpilot. Thanks for the PR!

PR #131, opened 2023-03-21
Update values.py to include 2021 Sorento FW. 

tonzy74:


sunnyhaibin:
Thanks for the PR. This should be opened against `master` at `commaai:openpilot`. If you can provide the dongle ID, I can help with making a PR for this.

PR #135, opened 2023-03-23
gpxd: remove secret

sunnyhaibin:
**Resolves**
- https://github.com/sunnyhaibin/sunnypilot/issues/133


tomhughes:
So that is now pulling the token from S3 if I'm reading that right? I'm not really sure how that helps if anybody call pull it?

tomhughes:
That still concerns me - the normal process would be for each user to authorise access and get their own token though obviously that would mean them using their own account.

Account sharing is also obviously a concern and I'm not aware of a case previously where people have tried to do it client side like this rather than via a server - do individual users have to opt in to having their traces uploaded or is it automatic? What do you do about getting their consent to our terms of use before uploading their data?

sunnyhaibin:
> That still concerns me - the normal process would be for each user to authorise access and get their own token though obviously that would mean them using their own account.
> 
> Account sharing is also obviously a concern and I'm not aware of a case previously where people have tried to do it client side like this rather than via a server - do individual users have to opt in to having their traces uploaded or is it automatic? What do you do about getting their consent to our terms of use before uploading their data?

Thank you for expressing your concerns about the use of sunnypilot software for uploading gpx traces to the OSM server. We understand that security and privacy are important.

To clarify, sunnypilot is a free software that users can choose to install on their devices. The gpx traces uploads are done automatically in the same way as openpilot uploads driving logs to the comma server. Users have full discretion on whether they want to install sunnypilot and upload their driving logs to remote server or not, including gpx traces.

sunnypilot can save users time and effort compared to other methods of contributing to OpenStreetMap. By automating the upload process, users do not have to manually create accounts/tokens to upload traces, which can be a discourage and time-consuming task to the majority of the users. This can make it easier and more convenient for users to contribute to OpenStreetMap.

As for getting user consent to your terms of use, we do not currently have a process for this. However, we would be willing to implement a consent mechanism if that is something you require.

tomhughes:
So is uploading logs the only thing sunnypilot does? I thought it was just one small thing in a much wider feature set - if the only reason to install it is because you want to upload your traces then obviously that is a bit different.

sunnyhaibin:
> So is uploading logs the only thing sunnypilot does? I thought it was just one small thing in a much wider feature set - if the only reason to install it is because you want to upload your traces then obviously that is a bit different.

sunnypilot is an open source driver assistance system that is installed on a device and makes use of the device's integrated GPS transmitter (with < 1m accuracy) to provide accurate positioning data. In addition to the ability to upload driving logs to OpenStreetMap, sunnypilot offers a range of features designed to improve the safety and efficiency of driving.

The ability to upload driving logs to OpenStreetMap is one small feature included in sunnypilot's wider feature set. However, we understand that security and privacy are important and we are committed to ensuring that users have full discretion over whether they want to install sunnypilot and use this feature or not. We are also open to implementing a consent mechanism if that is something you require.

tomhughes:
Right but to be clear right now everybody who chooses to install it is uploading logs automatically - it's not something they have to opt in to at all?

sunnyhaibin:
> Right but to be clear right now everybody who chooses to install it is uploading logs automatically - it's not something they have to opt in to at all?

Yes, by default sunnypilot uploads driving logs to our server automatically. However, users have the option to change the code and opt-out or re-opt-in if they desire. We acknowledge that having an opt-in mechanism would be a more user-friendly approach, and we are willing to implement this by adding on-screen toggles on the comma devices to make it easy for users to choose whether they want to upload their data or not.

PR #142, opened 2023-03-27
mapd: utilize advisory speed limit in curves

pfeiferj:
Whenever a curve has a marked advisory speed limit ignore the calculated speed for the curve and use the advisory speed. The advisory speed limit should be set on the way between the two curve hazard/speed advisory signs. The advisory speed limit should be set based off of the posted advisory speed for the curve. Using the advisory speed limit should be more accurate than our own speed calculations since osm doesn't have the ability to store true curves resulting in the calculations being inexact. The advisory speed limit should have been posted based off of an engineering study of the curve to provide a suggested speed.


sunnyhaibin:
@pfeiferj Just got around to testing this out. Seems to be working well in my limited testing. Are any changes still needed to this?

pfeiferj:
> @pfeiferj Just got around to testing this out. Seems to be working well in my limited testing. Are any changes still needed to this?

I believe it's good to go

sunnyhaibin:
Thanks for the PR! 🎉

PR #158, opened 2023-05-28
Subaru legacy: flip driver torque

sunnyhaibin:
This resolves permanent steering faults when the user overrides steering and then releases the steering wheel. Thanks to @martinl for this fix.
- https://github.com/martinl/panda/commit/8b7da38988a66b269a310479ed08973fb3635807

**Requires**
- https://github.com/sunnyhaibin/panda/pull/6


sunnyhaibin:
The user confirmed the issue is fixed. `b81cc3ea44acd372|2023-05-28--18-14-13`

PR #168, opened 2023-06-04
Update README.md | Update Recommended Link

Emalton:
The [1] marker in the README mentions an old link to an old branch, so I fixed that. It should use the new recommended branch.


Moodkiller:
Good find. I believe sunny was planning to unify the branch names so that links like this this don't have to be updated each release :). My suggestion, change the foot note to `smiskol.com/fork/sunnyhaibin/<branch-name>`

sunnyhaibin:
I have plans on updating the README regarding branch recommendations but will merge this PR as-is for now. Thanks for the PR!

PR #173, opened 2023-06-07
Ram DT: check EPS firmware explicitly to set minEnableSpeed to 0

sunnyhaibin:
The upstream method to check if the first 4 characters of the EPS firmware are smaller than or equal to the defined bounds to set `minEnableSpeed = 0.` would not work for some Ram DT.

Affected route: `001ccccec7a1992c|2023-06-07--21-25-45`
EPS FW: `b'21590101AA'`

Only allow `minEnableSpeed = 0.` for the EPS firmware that is known to allow enable steering at 0.


sunnyhaibin:
The affected users verified that the change allows the EPS to steer under the defined conditions.

PR #198, opened 2023-08-10
Nav WebServer Enhancement

mike8643:
Small enhancement to the non-prime navigation web interface with the following:
1. Ability to select existing home/week saved destinations that were stored in param ApiCache_NavDestinations from new dropdown
2. Ability to add 3 other favorite destinations that are stored in param ApiCache_NavDestinations with labels
3. On nav_confirmation template, added the word "Set" to separate action from main page. e.g. "Set Home", "Set Work" etc. 
osi1
osi2
osi3


sunnyhaibin:
Hmm, it seems like the PR was automatically closed after the branch was force pushed. Bad timing :(

Can you try to reopen again after rebasing?

PR #202, opened 2023-08-21
local nav server - added driving directions

mike8643:
**Description** [](A description of the refactor, including the goals it accomplishes.)
Added Navigation driving directions to the local nav web server. 
Saves a lightweight json version of the navd mapbox routing pulling out unnecessary components.
Adds to the json, Current Destination and Current Step.
New tpl for javascript to display information on the screen.
After entering destination through webserver, driving directions will immediately show up. 
If refreshed, webpage will display driving directions if param NavDestination is populated, in the case where the user enters the address directly onto the Comma device. 
Displays Current Destination, and clearer step text while removing previous steps once completed.
Refreshes every 5 seconds for new data.  
Known Limitations:
Does not decrease distance as you drive during that step, I did not want to add additional code to support this to make it lighter.
No dark mode, so the webpage is still using white background and black text. A bit jarring
Without Comma device in the onroad state, entering new destination will not refresh destination or directions since navd does not attempt to pull new mapbox api until device is in the onroad state. 
I am not a programmer by trade, this code works for me but can always be optimized. 
**Verification** [](Explain how you tested the refactor for regressions.)
Works fine on my end during my testing period of over a week. Used mobile iOS safari browser as well as chrome loaded onto my infotainment system. 
-->


sunnyhaibin:
Seems like force pushing got this automatically closed again... Will ping you and open another PR with this change once I get to this on `dev-c3` (cherry-picking should retain the commit history in the new PR). Thanks for the PR! 🎉

PR #204, opened 2023-08-25
Add Tucson Hybrid 2021 carFw

leZudiste:


leZudiste:
c3231807b9f5b5f6

sunnyhaibin:
Thanks for the PR. This should be opened against `master` at `commaai:openpilot`. If you can provide the dongle ID, I can help with making a PR for this.

PR #209, opened 2023-09-16
mapd: Bump `scipy` to 1.11.1

devtekve:
**Description** 
This commit updates the version of scipy in custom_dep.py from 1.7.1 to 1.11.1. This change is done to allow installing from non prebilt branches as the former scipy requested is incompatible with python 3.11.

**Verification** 
Installed from my custom branch after a device reset and install worked without errors of missing `overpy`  or missing `scipy`


sunnyhaibin:
Thanks for the PR! 🎉

PR #210, opened 2023-09-17
Adding fingerprint for palisade 2024 hda2

bjakiel:


sunnyhaibin:
Since we are mainly focusing on the PR in upstream, it isn't necessary to merge changes into the branch under the `sunnypilot` repo. The torque change is something that needs to be communicated with comma. Will close that PR and add the FW into the PR branch (and also the install branch under sunnypilot).
- https://github.com/commaai/openpilot/pull/27392

PR #213, opened 2023-09-20
HKG: Add FW versions for Kia K5 Hybrid 2022

hletrd:


dhimmel:
Noticed that the PR says no files changed. Has this commit been added by the merge?

sunnyhaibin:
> Noticed that the PR says no files changed. Has this commit been added by the merge?

This is likely due to the fingerprints already merged in [commaai/openpilot](https://github.com/commaai/openpilot). It should be available in `dev-c3` now.

PR #219, opened 2023-10-02
Breakout Ford F150 Lightning 

coffee-cake-isaac:
Ford F150 MK14 and Ford F150 Lightning share the same controls but have different ECUs (for now; this could change with CAN-FD fingerprinting working later), and have different weights.

I've added support to properly identify an F150 Lightning.

This still puts the vehicle into dashcam only mode, though I think it might be worth removing it from dashcam only.


coffee-cake-isaac:
Withhold any changes on this, actually. I need to validate this. I was having some issues with similar changes. I need to retest.

PR #236, opened 2023-11-04
Update values.py

dude241:
Changed steering threshold max to 400


sunnyhaibin:
This is not mergeable as non-tested actuator limit changes are against comma's safety policy. You can read more about it here:
- https://blog.comma.ai/understanding-the-openpilot-safety-model/
- https://github.com/commaai/openpilot/blob/master/docs/SAFETY.md

PR #240, opened 2023-11-11
Update README User Data section

gregorywaynepower:
Includes OpenStreetMaps as a place where users will expect their data to be shared in the form of GPS Traces in light of issue https://github.com/sunnyhaibin/sunnypilot/issues/201


sunnyhaibin:
Closing as the feature has been deprecated in all future releases of sunnypilot.

PR #242, opened 2023-11-15
Ford: CAN-FD speed limit

blue-genie:
Adding new functionality to feed Speed Limit that the car is aware of to SunnyPilot


blue-genie:
> How well is this functioning?

The change is already in SP dev-c3 branch.

PR #249, opened 2023-12-14
Update values.py

stevenctc:


sunnyhaibin:
This PR should be opened in commaai's openpilot. Once it's merged in upstream, we will merge the changes to add support for this firmware.

PR #252, opened 2023-12-16
fix: nnff require perfect match for car (not eps)

twilsonco:
requires exact match of car fingerprint in the NNLC model path, making it so that "fuzzy" matches are only allowed for the eps firmware string.


sunnyhaibin:
Will push out to the latest `dev-c3`. Thanks for the fix!

PR #256, opened 2024-01-02
Show Ford Gap Animation

blue-genie:
Display the Ford GAP UI when the GAP is adjusted.


crashnburn0:
I will be testing this on Q4 and Q3 fords. I will update the notes with the results.

sunnyhaibin:
@blue-genie Can you please rebase and point the PR against `master`?

blue-genie:
Let me do this in my branch, ensure that it works fine - then I'll rebase.

PR #258, opened 2024-01-11
NNLC adjustments and updated models

twilsonco:
* Improve NNLC runner to mellow out response with new models
  * When no "deliberate" lateral jerk is desired, lateral jerk is not used at all for feedforward or error correction
  * Decreased tolerance on what counts as "deliberate" lateral jerk
* Improve model lookup process
* New batch of community-data-based NNLC models
  * EPS-specific models where possible
  * Time-aligned input training data for higher quality fits


Moodkiller:
@twilsonco - any chance I can get you some logs / routes for a Kia Seltos 2023 non-SCC before this is merged? Or is it too late?

PR #264, opened 2024-01-19
torque: lateral jerk with stock torque controller

twilsonco:
I drove on this tonight. 
• Above very low speeds, it feels very similar to NNFF in my car. My car has very non-linear torque steering response at low speeds, so I wouldn't expect it to feel just like NNFF, but for Toyota/Hyundai/Subaru/Honda/Chrysler/Ram it should feel basically the same as NNFF, but it's much, much simpler. 
• No training models or data collection. 
• It takes my primary observations about the overall behavior of NNFF models. Seems to be the main benefit of NNFF is the limited use of lateral jerk for controls, so this tries to capture that.
• Works with and on top of torqued, so you still get live-updated controls (for better or worse 🙂 )

What it's missing compared to NNFF, is that this has no speed dependence whatsoever (problem for Volt and some other cars) and assumes very simple relationships between steer torque and lateral accel, lateral jerk, and road roll, since it's a minimal change to the torque controller. 

Compared to NNFF which learns the correct steer command for each and every combination of speed/lateral accel/lateral jerk/road roll. So NNFF knows that if the road is steeply crowned, it affects the correct steer torque at all speeds, path curvatures, etc., and that at every speed, the correct torque for a given curvature and a given road crown can change. 

Regardless, for Toyota/Hyundai/Subaru/Honda/Chrysler/Ram, my observation (from looking at plots, not from driving those cars) is that such nuances are nearly negligible.
This should go in the Lateral Tune section of the Controls settings, and include "(ALPHA FEATURE, PLEASE PROVIDE FEEDBACK)" type of disclaimer so that people are encouraged to report their findings.


twilsonco:
Can I get some feedback? Is this closed won’t implement? 

I wrote it so it’s gated behind a Boolean, so that it’s easy to gate with a toggle in settings. If this is similar in performance to NNLC for most cars, it will get is most of the way to what live-nnlc would provide, except arguably better since much more closely tied to the torque controller which has been exhaustively validated by comma.

PR #269, opened 2024-01-25
Sync: `commaai/openpilot:master` into `sunnypilot/sunnypilot:master`

devtekve:


devtekve:
This is currently working

lachezartodorov:
The driving works well in Chill and Experimental mode. NOO works too.

PR #270, opened 2024-01-26
NNLC don't need no freakin' lateral plan!

twilsonco:
This PR removes any dependence on the now-removed lateral plan.  NEEDS TESTING SINCE MY CAR IS CURRENTLY KAPUT!


devtekve:
This is still valid but got closed due to an update in dev-c3

PR #282, opened 2024-02-15
KIA K5 2024 HEV fingerprint

lssqla:


sunnyhaibin:
This is a car port PR, any reason why it is not opened with commaai's openpilot? This should be opened in commaai's openpilot. Once it's merged in upstream, we will merge the changes to add support for this car.

PR #308, opened 2024-03-20
NNFF hotfix for "lazy" steering in tight curves. Also optional disable of NNFF lateral jerk

twilsonco:


sunnyhaibin:
Merging this in the latest `dev-c3` with toggles exposed in settings after testing. Thanks!

PR #318, opened 2024-04-26
remove duplicate palisade 2023 line

yufeng66:


sunnyhaibin:
That's actually meant to be there as the default :)

PR #342, opened 2024-06-04
2018 Acura TLX - Serial Steering 

mlocoteta:
Adding TLX , verified on regular 9th gen accord and 2018 TLX. Not sure the years supported at this time. 

There was some bug fixes required in carstate.py. Specifically it looks like many models were renamed from something like CAR.CIVIC to CAR.HONDA_CIVIC. The only one which I wasn't quite sure about was CAR.ACCORDH - I couldn't find a replacement for this one. Everything else seemed pretty straight forward, some of the spacing changes are due to the checks before commit (unused variables/ line length > 160 characters). 

Todo: 
Need to generate docs.py.

- [\] added entry to CAR in selfdrive/car/*/values.py and ran `selfdrive/car/docs.py` to generate new docs
- [X] test route added to [routes.py](https://github.com/commaai/openpilot/blob/master/selfdrive/car/tests/routes.py)
- [X] route with openpilot: dc27c011224d353f/00000023--d15078f2fd/0
- [X] route with stock system: dc27c011224d353f/00000027--e5009d2033/0
- [X] car harness used (if comma doesn't sell it, put N/A): N/A - Serial Steering V3 board


mlocoteta:
There's a problem with the car docs- it looks like 2022 Hyundai Elantra is missing params, However, the TLX should have necessary params

mlocoteta:
Rebased to latest, fixed .gitmodules and what submodules point to.

sunnyhaibin:
@mlocoteta Can you please point this PR against `master`? Thanks!

selfdrive/car/honda/interface.py
```diff
@@ -173,7 +178,6 @@ def _get_params(ret, candidate, fingerprint, car_fw, experimental_long, docs):
       ret.lateralParams.torqueBP, ret.lateralParams.torqueV = [[0, 239], [0, 239]]
       ret.lateralTuning.pid.kiBP, ret.lateralTuning.pid.kpBP = [[0.,20], [0.,20]]
       ret.lateralTuning.pid.kpV, ret.lateralTuning.pid.kiV = [[0.4,0.3], [0,0]]
-      tire_stiffness_factor = 0.8467
```
sunnyhaibin, line 176: Let's remove this in a separate PR.

selfdrive/car/honda/interface.py
```diff
@@ -220,7 +224,6 @@ def _get_params(ret, candidate, fingerprint, car_fw, experimental_long, docs):
       else:
         ret.lateralParams.torqueBP, ret.lateralParams.torqueV = [[0, 2560], [0, 2560]]
         ret.lateralTuning.pid.kpV, ret.lateralTuning.pid.kiV = [[0.8], [0.24]]
-      tire_stiffness_factor = 1.
```
sunnyhaibin, line 223: Same as the above.

sunnyhaibin (changes requested):

sidbmw:
Which engine is this for and FWD or SH-AWD? Not sure if those make a difference. Might be worth adding trim info.

sunnyhaibin:

PR #346, opened 2024-06-15
Honda: Add Fit Hybrid 2018 JDM

AravindaDP:
**Checklist**

- [ ] added entry to CAR in selfdrive/car/*/values.py and ran `selfdrive/car/docs.py` to generate new docs
- [ ] test route added to [routes.py](https://github.com/commaai/openpilot/blob/master/selfdrive/car/tests/routes.py)
- [ ] route with openpilot:
- [ ] route with stock system:
- [x] car harness used (if comma doesn't sell it, put N/A): N/A (https://github.com/commaai/neo/pull/80)

Additional info:
Honda Fit Hybrid 2018 uses Nidec Connector similar to Clarity which has 3 CAN busses including RADAR CAN. And also, the same Brake message structure. However, unlike Clarity, all messages required for control is accessible through primary Vehicle CAN bus and extra CAN bus can be wired to passthrough in harness. See https://github.com/commaai/neo/pull/80

This is my first model port. Appreciate feedback on anything that needs to be done differently. (Specially params.toml updates etc)


sunnyhaibin:

PR #350, opened 2024-06-19
2018 Acura MDX Hybrid Car Port

mlocoteta:
**Car Port**

This was done AFTER the TLX was added, so this contains both in the diff until TLX is merged (or you can merge this instead)

[ ] added entry to CAR in selfdrive/car/*/values.py and ran `selfdrive/car/docs.py` to generate new docs 
- [x] test route added to [routes.py](https://github.com/commaai/openpilot/blob/master/selfdrive/car/tests/routes.py)
- [x] route with openpilot: c6c6b318abec90b8/00000016--2220084b30
- [x] route with stock system: c6c6b318abec90b8/00000017--88ede6eae6
- [x] car harness used (if comma doesn't sell it, put N/A): Serial Steering V3 Board

Notes: 
- Still have a problem generating docs.py due to Hyundai definitions. 
- Fixed routes HONDA.CIVIC --> HONDA.HONDA_CIVIC 
- Had to deal with ruff complaints about line length in carstate, line 256.


mlocoteta:
Need to fix turn signal

PR #358, opened 2024-07-11
Add (disabled) Virtual Torque Blending for Tesla

carleeno:
This PR adds Virtual Torque Blending logic to Tesla AP3 model 3/Y. It is currently disabled until a Toggle is added for it. See TODO in `carcontroller.py`

**Route:** `bbbf82d987d681bc/00000157--ad0c8e4303`

**TL;DR:** Enabling the Virtual Torque Blending toggle allows the user to steer while OP is engaged. It works by taking the user's steering torque and adjusts the angle output from OP before sending it to the car's EPAS, allowing the user to easily nudge the car within the lane, nudge it into exit ramps, and manually take turns, all without disengaging OP.

Torque blending strength can be tweaked by changing the multipliers in values.py, although keep in mind the latency and rate limit's effect on the strength's real-world effect, see below.

**Limitation:** The EPAS will not allow more than 2.5 Nm of torque from the user (except in very brief bursts) without faulting, in addition the control loop is a bit too slow to keep up with a fast manual steering jerk from the user. As a result, the virtual blending switches to a silent disengagement of lat control in some scenarios, automatically re-enabling lat control once the user's torque has decreased and the user's steering angle is again close to the OP commanded angle, or if more than 1 second elapses without user torque even if there is still an angular disagreement. This switchover from torque blending to full override sometimes results in feeling a bump in the steering, similar to the bump when disengaging from classic AP. Work is still being done to try to minimize the strength of this bump, it's typically felt when turning sharp corners at low speeds.

**Safety considerations:**
- Ability to always override: this is baked in to the Tesla's EPAS, it's always possible to override the steering with ~ 2.5 Nm of force no matter what OP tries to do.
- Never silently release control to the user without the user having hands on the wheel: the main criteria for silently disengaging is to reach hands-on-level 3, which is a signal from the EPAS that roughly equates to > 2.5 Nm of force (with some filtering)
- Avoid accidental torque blending: there is a 0.5 Nm deadzone where the angle is not modified, so that a misbalance of the wheel, or slight hand weight, will not cause it to veer from OP's lat control.
- Avoid high-rate steering even if data is faulty: the steering rate limiting is applied *after* the torque blending, so even if OP receives faulty torque data, the rate limits will still be respected, giving the user time to override.
- Blending is disabled by default behind a toggle to ensure the user is aware of the feature, reducing risk of accidental nudging from misunderstanding the capability.

As always, it's possible not all safety considerations were realized, or that an edge case was missed, use with caution.


carleeno:
Github didn't like it when I changed the base, I'll spin a new PR shortly.

PR #359, opened 2024-07-17
Honda Longitudinal: Remove tuning from 4-CYL 9th Gen Accord

mlocoteta:
Updated tuning to use the standard Honda tuning. No longer needed on both pedal and PCM control.


sunnyhaibin (approved):

sunnyhaibin:
Thanks for the PR!

PR #362, opened 2024-07-29
MADS: Hyundai: Independent control with LFA button

sunnyhaibin:


sunnyhaibin:
Confirmed the button states are working as described by the control policy.

PR #363, opened 2024-07-30
MADS: Honda: Fix PCM ACC disengage fault

sunnyhaibin:


sunnyhaibin:
Verified by user with a confirmed fix on the bug.

devtekve (approved):

PR #367, opened 2024-07-31
[SL] Improving authentication mechanism and help with local dev

devtekve:
Changing from using cookies to headers for the authentication and adding conditional requirement of SSL based on whether this is localhost or not.


sunnyhaibin:
@devtekve Ready to merge?

devtekve:
@sunnyhaibin let's bring it to dev-c3 for a little while

sunnyhaibin:
Move to Draft until it's tested on `dev-c3` for a few.

sunnyhaibin:
@devtekve is this still relevant?

PR #369, opened 2024-07-31
Remove dangling imports

carleeno:
SP failed to start due to an import error on button states. I noticed create_mads_event was also unused.

This gets the tesla port to a working state as long as mads is disabled.


sunnyhaibin:
Thanks!

PR #376, opened 2024-08-01
Tesla: Add speed limit detection

carleeno:
This adds car-based speed limit detection for Tesla.

This needs to be tested by someone in a KPH region, as I'm not certain the DI speed units applies to the DAS fused speed limit signal. the DAS frame does not have it's own units signal. I was only able to confirm that it's not based on the UI speed units signal.

**Route:** `bbbf82d987d681bc/00000239--f1dcd4a66a/0`


sunnyhaibin:
Can you reopen this PR against the master-new branch at https://github.com/sunnypilot/opendbc?

PR #378, opened 2024-08-01
car: Move all live param checks to `ParamManger`

sunnyhaibin:
Part of https://github.com/sunnypilot/sunnypilot/issues/377


sunnyhaibin:
Confirmed working for all params in `ParamManager`.

PR #382, opened 2024-08-02
car: Fix typo with `ParamManager` variable

sunnyhaibin:
Introduced in https://github.com/sunnypilot/sunnypilot/pull/378


sunnyhaibin:
Fix confirmed by user.

PR #384, opened 2024-08-03
Tesla: some MADS fixes

carleeno:
This gets MADS to a sorta-working but still flakey state

Biggest 2 issues are:
- half-press down sometimes doesn't prevent ACC from engaging
- full-press up to turn off MADS results in shifting to reverse if below 5mph!!

`bbbf82d987d681bc/00000008--d548c019f5`

### DIFF
--- a/selfdrive/car/tesla/carcontroller.py
+++ b/selfdrive/car/tesla/carcontroller.py
@@ -53,6 +53,7 @@ class CarController(CarControllerBase):
     if self.frame % 10 == 0 and pcm_cancel_cmd and CS.acc_enabled:
       counter = int(CS.sccm_right_stalk_counter)
       can_sends.append(self.tesla_can.right_stalk_press((counter + 1) % 16 , 1))  # half up (cancel acc)
+      can_sends.append(self.tesla_can.right_stalk_press((counter + 1) % 16 , 0))  # back to 0 for falling edge detection
 
     # TODO: HUD control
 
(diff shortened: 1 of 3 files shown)


carleeno:
From discord:

**observed stalk quirks with MADS:**

half-press down while too slow for acc still engages MADS, but car alerts that "Cruise is unavailable"

full-press down from nothing results in both cruise and lat, so there's currently no cruise-only option

half-press up when cruise isn't enabled results in "Hold for neutral" message, but this isn't a problem because you would only half-press up to cancel cruise as it's currently programmed

full-press up to turn off mads when acc is already off results in an attempt to shift to reverse. If below 5mph and turning off mads, the car actually goes into reverse.

The first 3 are only annoyances, the last one is serious, I think we can't use full-up for anything, we shouldn't encourage that input to be used regularly.

also it seems flakey that sometimes half-down results in lat+acc, and I can't tell why. It seems when that happens we don't send the half-up to cancel acc, which tells me we're not getting the CC.cruiseControl.cancel signal.

**Input alternatives I'm thinking of after playing with it:**

A: stalk press counter: 1x, 2x, 3x for mads, both, acc (or acc, both, mads based on user preference). Only use half up once to turn everything off, avoiding reverse issue. mads or acc can individually be turned off by steering or brake overrides respectively.

B: or keep current half/full down logic, but add falling edge detection with a max() to allow for "full-press-only", and use half-up to turn off everything at once, avoiding reverse issue.

C: or half-down for acc, full-down for acc+lat, half-up for mads on (if everything is off)...or half-up for everything off (if anything is on)

D: or 1x down for acc, 2x down for acc+lat, half-up for mads on (from off state)...or half-up for everything off (from engaged state)

C-D result in the fewest alerts, but still get an alert for enabling mads, maybe acceptable. However having half-up do 2 different things based on state might be confusing.

A and D are the closest to OEM Tesla behavior, and right now A is the best sounding option to me to be intuitive for existing Tesla owners upgrading to SP. maybe a bit weird for existing SP users downgrading to a tesla though 😉

carleeno:
will be superseded shortly

PR #387, opened 2024-08-03
Docs: Move HOW-TOS to `master`

devtekve:
I found that because the branch `(!)README` is so far behind, it causes GIT to store a lot of granular commit info separated +  all the git history before the branch was cleared which causes every  `git fetch` to waste a whopping 1.08GiB! 

I've moved the only thing that was missing which is the `HOW-TOS.md` to master branch and updated the `README.md` accordingly

image

After this is merged, we'll nuke the branch `(!)README`


sunnyhaibin (approved):
LGTM. We will refactor this with the new docs once it's ready for live.

PR #388, opened 2024-08-03
ui: Updater: More verbose when running

devtekve:
This should help people to feel less frustrated when they tap on "check" and it just goes grey and it just seems like hanging in there...


sunnyhaibin:
@devtekve Is this still relevant? Can you rebase with `master-new`?

PR #389, opened 2024-08-03
Refactor update process execution and signaling

devtekve:
Changed update process execution from a Python to a native process. Updated signal-based process control to match new process configurations.

**Description**

https://github.com/commaai/openpilot/pull/32716 removed the process title, which meant that we could no longer send signals to it as we used to which caused this issue https://github.com/commaai/openpilot/issues/32881. 

Following recommendations here https://github.com/commaai/openpilot/pull/32716#issuecomment-2166919909, we are updating it to be a native process and update the signal sent from software_settings

**Verification**

Installed this branch on my devices and clicked the update button and I saw on the logs the signal was received.


devtekve:
This has been pushed to `master-dev-c3`

sunnyhaibin:
This is supported in the latest `master-new`.

PR #398, opened 2024-08-06
Nnlc adjustable lat jerk

twilsonco:
Simple change to NNLC and `TorqueLateralJerk`, replacing `NNFFNoLateralJerk` with `NNFFLateralJerkFactor`, i.e. making it an adjustable float rather than a bool. This will allow for better fine-tuning of lateral jerk NNLC and "TorqueLateralJerk" input.

Also includes two bugfixes, 

1. the `lat_accel_friction_factor` would be increased when there's no desired lateral jerk, but it wasn't being lowered again in case there was non-zero desired lateral jerk. This would have resulted in too-high "friction compensation" after the bug was tripped. 
2. Because `self.use_lateral_jerk` can be changed onroad, it was possible to trigger a crash due to some of the lateral jerk instance variables having not been defined inside `init()`. The fix is to define all potentially necessary instance variables in `init()` regardless of NNLC and TorqueLateralJerk settings.

Both of these changes (first bugfix and adjustable lateral jerk) should help folks experiencing overactive NNLC and "TorqueLateralJerk" steering.

### DIFF
--- a/selfdrive/controls/lib/latcontrol_torque.py
+++ b/selfdrive/controls/lib/latcontrol_torque.py
@@ -83,8 +83,7 @@ class LatControlTorque(LatControl):
     self.torqued_override = self.param_s.get_bool("TorquedOverride")
     self._frame = 0
 
-    self.use_lateral_jerk = self.param_s.get_bool("TorqueLateralJerk")  # TODO: make this a parameter in the UI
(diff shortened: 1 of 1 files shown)


sunnyhaibin:
Thanks! :heart:

PR #410, opened 2024-08-13
Show GAP UI Animation

blue-genie:
continuation of #256


blue-genie:
@sunnyhaibin This is the new PR for Ford GAP Animation. We need to keep sending the "send_ui" flag as long as we want the animation to be displayed, in my code I'm keeping it for 4 seconds.

sunnyhaibin:

PR #421, opened 2024-08-28
check if tranmissionType is unknown in case of a manual Honda where g…

lx93:
On the 6MT 2018-2020 Honda Fit, gearbox does not show up on CAN, therefore I updated the can_parser to check for these situations and to not expect GEARBOX message. Also changed so the GearShifter is in Drive when transmissionType is unknown.

**Car**
2018 Honda Fit Ex Manual

**Route**
27f27145758a513f/000000d6--43b1cf15e0/0


sunnyhaibin:

PR #429, opened 2024-09-20
Use escc alternatively

jcwlim:
Use ESCC lead data in alternative distance calculation.
Default set to skip every 1m.

This effectively reduce the sudden brake and acceleration during motor cycle cut in with fast speed.
It does still happen but this reduce 80% - 90% of it.


devtekve:
@jcwlim this is not forgotten, we are refactoring a ton. I will get back to you once we've refactored so we can update your PR to the latest code (I will help you updating it <3 )

PR #430, opened 2024-09-27
Bring the good back to LX RX350 LSS2

Kenstea:
Undo the change that causes the slow acceleration for LSS2 on RX350.

**Description**

Undo the change that causes the slow acceleration for LSS2 on RX350.

**Verification**

The slowness occurs after installing the change.


Stupefacient:
Please rebase this to only include your single commit and not other older commits

Kenstea:
> Please rebase this to only include your single commit and not other older commits

The base branch has undone some code. Close this PR as it based off the older one.

PR #433, opened 2024-09-28
Driving Model Selector v6

sunnyhaibin:


mikehundley:
For what it's worth, I ran this commit on my can-fd Hyundai Tucson 2022 and it would not calibrate. Stayed at 0% on standard models as well as the new ones (TR, GB). I attempted swapping from master after calibration and received a radar error when engaging. Curiously it looks like lateral (MADS) does work however I did not try this in depth.

PR #446, opened 2024-11-15
Modular Assistive Driving System (MADS)

devtekve:
**TODOs**
- [x] SET/- button does not emit correct event if cruise not allowed in openpilot longitudinal
- [x] tests for MADS state machine

Relates to: 
* Panda PR:
  * https://github.com/sunnypilot/panda/pull/40
  * https://github.com/sunnypilot/panda/pull/50
  * https://github.com/sunnypilot/panda/pull/57
* Opendbc PR: https://github.com/sunnypilot/opendbc/pull/29

## Summary by Sourcery

New Features:
- Introduce MADS to customize the activation and deactivation of lateral and longitudinal control.


sunnypilot/selfdrive/car/hyundai/__init__.py
devtekve, line 1: For make-specific configs I'd prefer to have them similar to a change being introduced for the ESCC where things reside on `opendbc` instead of parts of it on the openpilot repo. 

https://github.com/sunnypilot/opendbc/pull/28/files#diff-c91792bcc46f80b658bcf316682d700e8d3c74669bd5a83b2f8169f7c4247190

The reason for this is that make-specific changes will happen always on opendbc, and this will help avoid a situation like we currently have on the PR for the `opendb`  https://github.com/sunnypilot/opendbc/pull/29/files#diff-b0e7d06078160efd89cb9f7ab7b991c90eed01318b48fc0b0f0987e8d136eebf where from that PR you can't tell what's really happening on CarControllerSP for hyundai.

Another argument for this is that objectively speaking, SP repo depends on opendbc submodule. But opendbc submodule now also depends on SP repo, introducing a (logical) circular dependency. 

If these were 2 separate packages that you could install, it would be unable to, because SP depends on Opendbc and Opendbc now depends on SP and so on...

cereal/log.capnp
```diff
@@ -125,6 +125,76 @@ struct OnroadEvent @0xc4fa6047f024e718 {
...
+    eventReserved141 @141;
+    eventReserved142 @142;
+    eventReserved143 @143;
+    eventReserved144 @144;
+    eventReserved145 @145;
+    eventReserved146 @146;
+    eventReserved147 @147;
+    eventReserved148 @148;
+    eventReserved149 @149;
+    eventReserved150 @150;
```
devtekve, line 185: How do we guarantee that these ordinals won't be taken in the future? shouldn't we create our own event in custom instead?

sunnyhaibin: These are reserved for comma’s expansions, not sunnypilot’s.

devtekve: Oh, so you are pre allocating them and whenever they come up with new evetns we just "sacrifice" one of thse reserve and thats how we can be on a higher ordinal without impact?

sunnyhaibin: Yes.

selfdrive/ui/ui.cc
```diff
@@ -68,13 +68,26 @@ void ui_update_params(UIState *s) {
 }
 
 void UIState::updateStatus() {
-  if (scene.started && sm->updated("selfdriveState")) {
+  if (scene.started && (sm->updated("selfdriveState") || sm->updated("selfdriveStateSP"))) {
     auto ss = (*sm)["selfdriveState"].getSelfdriveState();
+    auto ss_sp = (*sm)["selfdriveStateSP"].getSelfdriveStateSP();
+    auto mads = ss_sp.getMads();
```
sunnyhaibin, line 74:
```suggestion
    auto mads = (*sm)["selfdriveStateSP"].getSelfdriveStateSP().getMads();
```

sunnypilot/mads/helpers.py
```diff
@@ -0,0 +1,71 @@
...
+from openpilot.common.params import Params
+
+from opendbc.car.hyundai.values import HyundaiFlags
+from opendbc.sunnypilot.car.hyundai.values import HyundaiFlagsSP
+
+
+class MadsParams:
+  def __init__(self):
+    self.params = Params()
+    self.allow_always = False
```
sunnyhaibin, line 38:
```suggestion
```

cereal/log.capnp
```diff
@@ -125,6 +125,82 @@ struct OnroadEvent @0xc4fa6047f024e718 {
...
+    silentLkasDisable @157;
+    silentBrakeHold @158;
+    silentWrongGear @159;
+    silentReverseGear @160;
+    silentDoorOpen @161;
+    silentSeatbeltNotLatched @162;
+    silentParkBrake @163;
+    controlsMismatchLateral @164;
+    wrongCarModeNoEntry @165;
+    wrongCruiseModeNoEntry @166;
```
sunnyhaibin, line 203: Could be removed if no longer in use.

```suggestion
```

scripts/lint/lint.sh
```diff
@@ -57,7 +57,7 @@ function run_tests() {
 
   if [[ -z "$FAST" ]]; then
     run "mypy" mypy $PYTHON_FILES
-    run "codespell" codespell $ALL_FILES
+    run "codespell" codespell $ALL_FILES -L Wen
```
sunnyhaibin, line 60: Do it like https://github.com/sunnypilot/opendbc/commit/edba9e8fb6576a0f9e5a5a4f225fc78f45758359

```suggestion
    run "codespell" codespell $ALL_FILES
```

selfdrive/car/card.py
```diff
@@ -113,6 +115,12 @@ def __init__(self, CI=None, RI=None) -> None:
     if not disengage_on_accelerator:
       self.CP.alternativeExperience |= ALTERNATIVE_EXPERIENCE.DISABLE_DISENGAGE_ON_GAS
 
+    # mads
+    data_services = list(self.sm.data.keys()) + ['selfdriveStateSP']
+    self.sm = messaging.SubMaster(data_services, poll='selfdriveStateSP')
```
sunnyhaibin, line 120: Is this really needed in card?

selfdrive/selfdrived/events.py
```diff
@@ -951,6 +969,113 @@ def personality_changed_alert(CP: car.CarParams, CS: car.CarState, sm: messaging
...
+    ET.NO_ENTRY: NoEntryAlert("Controls Mismatch: Lateral"),
+  },
+
+  EventName.wrongCarModeNoEntry: {
+    ET.NO_ENTRY: wrong_car_mode_alert,
+  },
+
+  EventName.wrongCruiseModeNoEntry: {
+    ET.NO_ENTRY: NoEntryAlert("Adaptive Cruise Disabled"),
+  },
```
sunnyhaibin, line 1077: Remove if no longer in use.

```suggestion
```

selfdrive/car/card.py
sunnyhaibin, line 120:
```suggestion
```

devtekve:
FYI cross-posting comment https://github.com/sunnypilot/opendbc/pull/29#issuecomment-2544912728

selfdrive/ui/translations/main_es.ts
```diff
@@ -123,6 +123,26 @@
         <source>Longitudinal Maneuver Mode</source>
         <translation>Modo de maniobra longitudinal</translation>
     </message>
+    <message>
+        <source>MADS: Disengage Lateral on Brake</source>
+        <translation>MADS: Deshabilitar control lateral en frenada</translation>
+    </message>
+    <message>
+        <source>Modular Assistive Driving System (MADS)</source>
+        <translation type="unfinished"></translation>
```
devtekve, line 132: not gonna rename it in Spanish :D

selfdrive/selfdrived/events.py
devtekve: When and how were they meant to be used?

selfdrive/ui/ui.h
```diff
@@ -42,12 +42,16 @@ typedef enum UIStatus {
...
+  STATUS_LAT_ONLY,
+  STATUS_LONG_ONLY,
 } UIStatus;
 
 const QColor bg_colors [] = {
   [STATUS_DISENGAGED] = QColor(0x17, 0x33, 0x49, 0xc8),
   [STATUS_OVERRIDE] = QColor(0x91, 0x9b, 0x95, 0xf1),
   [STATUS_ENGAGED] = QColor(0x17, 0x86, 0x44, 0xf1),
+  [STATUS_LAT_ONLY] = QColor(0x00, 0xc8, 0xc8, 0xf1),
+  [STATUS_LONG_ONLY] = QColor(0x96, 0x1C, 0xA8, 0xf1),
```
devtekve, line 54:
```suggestion
  [STATUS_LONG_ONLY] = QColor(0xFF, 0x1C, 0xE3, 0xF1),
```
from:
image

to:
image

selfdrive/controls/controlsd.py
```diff
@@ -88,7 +92,16 @@ def state_control(self):
...
     standstill = abs(CS.vEgo) <= max(self.CP.minSteerSpeed, MIN_LATERAL_CONTROL_SPEED) or CS.standstill
-    CC.latActive = self.sm['selfdriveState'].active and not CS.steerFaultTemporary and not CS.steerFaultPermanent and not standstill
+
+    ss_sp = self.sm['selfdriveStateSP']
+    CC.madsEnabled = ss_sp.mads.enabled
+    if ss_sp.mads.available:
+      CC.sunnypilotParams |= SunnypilotParamFlags.ENABLE_MADS.value
+      _lat_active = ss_sp.mads.active
+    else:
+      _lat_active = self.sm['selfdriveState'].active
```
devtekve, line 102: Would it be possible to move this away of this file into something like `mads_get_lat_active()` [similar to what we have in panda](https://github.com/sunnypilot/panda/blob/master-new/board/safety.h#L117)? (_Just the abstraction, not the logic of overriding mads lat with controls allowed as we intentionally do it differently here_)

selfdrive/controls/controlsd.py
```diff
@@ -56,6 +57,9 @@ def __init__(self) -> None:
     elif self.CP.lateralTuning.which() == 'torque':
       self.LaC = LatControlTorque(self.CP, self.CI)
 
+    data_services = list(self.sm.data.keys()) + ['selfdriveStateSP']
+    self.sm = messaging.SubMaster(data_services, poll='selfdriveState')
```
devtekve, line 61: for my own knowledge, what does this line do exactly? I've never seen/used the `poll` myself

selfdrive/controls/controlsd.py
```diff
@@ -56,6 +57,9 @@ def __init__(self) -> None:
     elif self.CP.lateralTuning.which() == 'torque':
       self.LaC = LatControlTorque(self.CP, self.CI)
 
+    data_services = list(self.sm.data.keys()) + ['selfdriveStateSP']
```
devtekve, line 60: It kinda feels like we are injecting ourselves "late" within the "flow". Should perhaps we be earlier were the sm was instantiated?

selfdrive/car/card.py
```diff
@@ -113,6 +115,10 @@ def __init__(self, CI=None, RI=None) -> None:
     if not disengage_on_accelerator:
       self.CP.alternativeExperience |= ALTERNATIVE_EXPERIENCE.DISABLE_DISENGAGE_ON_GAS
 
+    # mads
+    MadsParams().set_alternative_experience(self.CP)
+    MadsParams().set_car_specific_params(self.CP)
```
devtekve, line 120: Would it be possible to maybe add these as part of the ctor of `MadsParams()`? would that make sense?

devtekve: something like 
```py
MadsParams(self.CP)
```

sunnyhaibin: While doable, `set_alternative_experience` and `set_car_specific_params` are both static methods meant to be used independently of any instance state. We don't necessarily need to instantiating the class for these since we don't need to create a new object.

selfdrive/controls/controlsd.py
sunnyhaibin: This happens at init, so it does not matter where this is handled as long as it's contained within init.

selfdrive/controls/controlsd.py
sunnyhaibin: `poll` sets the update frequency inside of `SubMaster`, which for `selfdriveState` is at 100 Hz.

selfdrive/ui/qt/offroad/developer_panel.cc
```diff
@@ -24,6 +24,20 @@ DeveloperPanel::DeveloperPanel(SettingsWindow *parent) : ListWidget(parent) {
   });
   addItem(longManeuverToggle);
 
+  // FIXME-SP: Move to sunnypilot panels before merging
```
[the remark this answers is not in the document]
sunnyhaibin: Dependent on `sp-ui`.

selfdrive/ui/ui.h
sunnyhaibin: `QColor` with 4 arguments is ARGB. This is the colors with the current hex assignments:
output

selfdrive/car/card.py
devtekve: `MadsParams().set_car_params(self.CP)` ?

selfdrive/controls/controlsd.py
devtekve: I think it would be better if it happens alongside the other SM added for this

selfdrive/ui/ui.cc
```diff
@@ -68,13 +68,25 @@ void ui_update_params(UIState *s) {
...
-      status = ss.getEnabled() ? STATUS_ENGAGED : STATUS_DISENGAGED;
+      if (mads.getAvailable()) {
+        if (mads.getEnabled()) {
+          status = ss.getEnabled() ? STATUS_ENGAGED : STATUS_LAT_ONLY;
+        } else {
+          status = STATUS_DISENGAGED;
+        }
+      } else {
+        status = ss.getEnabled() ? STATUS_ENGAGED : STATUS_DISENGAGED;
+      }
```
sunnyhaibin, line 89: TODO: Abstract MADS related logic.

selfdrive/car/card.py
sunnyhaibin: Let's revisit in another PR. Will add TODO-SP.

rav4kumar (dismissed):
lgtm

sunnyhaibin:
image

sunnyhaibin (approved):

PR #449, opened 2024-11-16
ci: sunnypilot show UI diff on PRs 

devtekve:
adds the branch we are currently using as one of the targets to trigger the action of diff


sunnyhaibin (approved):
LGTM!

PR #456, opened 2024-11-22
Bumping msgq to latest master

devtekve:


sunnyhaibin:
Superseded with latest sync.

PR #467, opened 2024-12-08
LFS: Sync refs from upstream

devtekve:
Used to pull comma's LFS and have it downloaded on your PC so that it's then ready to be pushed to our own LFS.


.github/workflows/selfdrive_tests.yaml
```diff
@@ -355,6 +355,7 @@ jobs:
         ${{ env.RUN }} "source selfdrive/test/setup_xvfb.sh && \
                         source selfdrive/test/setup_vsound.sh && \
                         CI=1 pytest tools/sim/tests/test_metadrive_bridge.py"
+      continue-on-error: true
```
sunnyhaibin, line 358: Handle in a separate PR, if it's still an issue.

```suggestion
```

sunnyhaibin, line 358: Handle in a separate PR, if it's still an issue.

```suggestion
```

sunnyhaibin (approved):

PR #470, opened 2024-12-08
ci: Skip simulator failures

devtekve:
Allows fails for the simulator test because we don't trust it, it fails even without changes


sunnyhaibin (approved):
Send it.

PR #473, opened 2024-12-09
Update values.py

krskubisz:
support for 2024 mustang gt


hiimisaac:
2024 Mustangs are encrypted. Even if we add this, we won't be able to send/receive anything on the CANBUS.

PR #476, opened 2024-12-09
CI: Build prebuilts and models from github and self hosted runners

devtekve:
Good working


.github/workflows/sunnypilot-release.yaml
```diff
@@ -0,0 +1,249 @@
+name: SunnyPilot CI
```
devtekve, line 1:
```suggestion
name: sunnypilot CI
```

PR #482, opened 2024-12-16
NNLC: use substitute.toml for fallback nnlc lookup

twilsonco:
This enables NNLC on many more cars based on Comma's substitute.toml file. If the user's car has a direct match, that model is used, or it looks for a substitute and loads a matching model for the sub. No UI changes are necessary. If a substitute match is made, the user will know because the model name in settings will not match their car.

Comma uses the substitute.toml file to load one car's torque parameters in place of another car. This change is based on the assumption that if that works for torque parameters, it should work for NNLC models. 

Tested with @smarf's '16 Outback, which has no NNLC model and is in substitute.toml pointing to `SUBARU_IMPREZA` which does have a NNLC model. The Impreza model loaded correctly and here's the initial feedback: 

CleanShot 2024-12-16 at 13 51 39


twilsonco:
Good to point that out @zorrobyte. Even more than with exact matches, YMMV with NNLC if a substitute match is used.

Even worse is that for some models, multiple generations are squashed into a single platform on OP despite differences in EPS. Chevy Bolt is the worst that I know of. I think this is an example of Comma choosing form over function, even though they've designed a system that allows for an arbitrary number of platforms that could easily handle all such variations.

If Comma would add in all these variations, then offline controls solutions like the current NNLC (including Comma's own NNLC implementation) could be more complete, hindered only by the availability of data to train models. Their approach (I speculate) is based on the assumption that torqued can adapt to fill the gaps between cars that have been squashed to a common platform. For NNLC, that means that the only reasonably complete solution is online training of NNLC models as you drive. Of course, that is the ideal solution regardless of how granularly Comma organizes platforms.

In the meantime, hopefully this will spread more of the NNLC love until online NNLC becomes a thing.

sunnyhaibin:
Will mark this PR as draft for now until we have reimplemented NNLC in the latest `master-new` at https://github.com/sunnypilot/opendbc.

sunnyhaibin:
@twilsonco will reopen.

PR #524, opened 2025-01-05
ui: prevent driving model change with offroad transition

devtekve:
## Summary by Sourcery

Bug Fixes:
- Fixed an issue where the selected model would be unintentionally changed when transitioning from offroad to onroad.


sunnyhaibin (approved):
LGTM

PR #536, opened 2025-01-08
WIP: Toyota Security Key Manager

calvinpark:
**Description**

Add Toyota Security Key Manager which allows for

1. Extracting the key from the car using extract.py that's based on https://github.com/I-CAN-hack/secoc/blob/main/extract_keys.py, and then installing the key.
2. Manually installing the key using a keyboard.

**Verification**

Manually tested a few different ways.

## Summary by Sourcery

New Features:
- Added Toyota Security Key Manager to allow extracting the key from the car and installing it manually.

### DIFF
--- a/selfdrive/ui/qt/offroad/settings.cc
+++ b/selfdrive/ui/qt/offroad/settings.cc
@@ -205,6 +205,18 @@ DevicePanel::DevicePanel(SettingsWindow *parent) : ListWidget(parent) {
 
   // offroad-only buttons
 
+  auto tskBtn = new ButtonControl(tr("Toyota Security Key Manager"), tr("RUN"));
+  connect(tskBtn, &ButtonControl::clicked, [this, tskBtn]() {
+    if (ConfirmationDialog::confirm(tr("Run Toyota Security Key Manager and reboot?"), tr("RUN"), this)) {
+      tskBtn->setEnabled(false);
+      QProcess *process = new QProcess(this);
+      // This shell script launches TSK Manager on top of OP and then reboots,
+      // so there's no need to handle hiding and restoring OP application.
+      process->start("/data/openpilot/tsk/manager.sh");
+    }
+  });
+  addItem(tskBtn);
+
   auto dcamBtn = new ButtonControl(tr("Driver Camera"), tr("PREVIEW"),
                                    tr("Preview the driver facing camera to ensure that driver monitoring has good visibility. (vehicle must be off)"));
   connect(dcamBtn, &ButtonControl::clicked, [this, dcamBtn]() {
--- a/system/ui/raylib/tsk-kbd.cc
+++ b/system/ui/raylib/tsk-kbd.cc
@@ -1,0 +1,329 @@
+#include "system/ui/raylib/util.h"
+#include <string>
+#include <vector>
+#include <fstream>
+#include <regex>
+#include <chrono>
+#include <algorithm>
+#include <cerrno>
+#include <cstring>
+
+// --- Constants ---
+const int FONT_SIZE = 100;
+const int INSTALLED_LABEL_FONT_SIZE = 80;
+const int ERROR_LABEL_FONT_SIZE = 80;
+const float FONT_SPACING = 1.0f;
+const float INPUT_FONT_SPACING = 3.0f;
+const Color GRAY = {96, 96, 96, 255};
+const int EXIT_BUTTON_PADDING = 20;
+const int INPUT_BOX_PADDING = 20;
+const int CHARS_LEFT_LABEL_PADDING = 20;
+const int KEY_HEIGHT = 180;
+const int KEY_PADDING = 10;
+const int NUM_KEYS_FIRST_ROW = 10;
+const int NUM_KEYS_SECOND_ROW = 7;
+const int INPUT_BOX_CHARS = 32;
+
+// --- Function to convert hex string to Color ---
+Color HexToColor(const std::string& hex) {
+    if (hex.length() != 7 || hex[0] != '#') {
+        return GRAY; // Default to gray if invalid
+    }
+
+    unsigned int r, g, b;
+    if (sscanf(hex.c_str(), "#%02x%02x%02x", &r, &g, &b) != 3) {
+        return GRAY; // Default to gray if invalid
+    }
+
+    return {static_cast<unsigned char>(r), static_cast<unsigned char>(g), static_cast<unsigned char>(b), 255};
+}
+
+
+// --- Function to write content to a file and capture error messages ---
+std::pair<bool, std::vector<std::string>> writeToFileWithError(const std::string& path, const std::string& content) {
+    std::ofstream file(path, std::ios::out | std::ios::trunc);
+    std::vector<std::string> errorMessages;
+    if (!file.is_open()) {
+        errorMessages.push_back("Failed to open file '" + path + "'");
+        errorMessages.push_back("Error: " + std::string(strerror(errno)));
+        return {false, errorMessages};
+    }
+    file << content;
+    file.close();
+    if (!file.good()) {
+        errorMessages.push_back("Failed to write to file '" + path + "'");
+        errorMessages.push_back("Error: " + std::string(strerror(errno)));
+        return {false, errorMessages};
+    }
+    return {true, {}};
+}
+
+// --- Function to read and validate the SecOCKey ---
+std::string readSecOCKey() {
+    const std::string filePath = "/data/params/d/SecOCKey";
+    std::ifstream file(filePath, std::ios::binary | std::ios::ate);
+
+    if (!file.is_open()) {
+        return "Installed: None";
+    }
+
+    std::streamsize size = file.tellg();
+    file.seekg(0, std::ios::beg);
+    std::vector<char> buffer(size);
+    if (!file.read(buffer.data(), size)) {
+        return "Installed: None";
+    }
+
+    if (std::any_of(buffer.begin(), buffer.end(), [](char c) { return c < 32 && c != '\n' && c != '\r'; })) {
+        return "Installed: Invalid (binary file)";
+    }
+
+    std::string content(buffer.begin(), buffer.end());
+    content.erase(std::remove(content.begin(), content.end(), '\n'), content.end());
+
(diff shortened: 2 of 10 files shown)


calvinpark:
WIP because the screen timeout is activating while OP is hidden behind the manager

tsk/extract.py
```diff
@@ -0,0 +1,298 @@
...
+  def __str__(self) -> str:
+    return f"{self.message}\n\n!!!! Try again. If the problem persists, turn off the car, put it back into 'Not Ready to Drive' mode, and then try again.\n"
+
+
+class TSKExtractor:
+  ADDR = 0x7a1
+  DEBUG = False
+  BUS = 0
+
+  SEED_KEY_SECRET = b'\xf0\x5f\x36\xb7\xd7\x8c\x03\xe2\x4a\xb4\xfa\xef\x2a\x57\xd0\x44'
```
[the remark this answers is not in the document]
calvinpark: This is a part of the exploit which should not be changed. It does not pose a security risk.

system/ui/raylib/tsk-kbd.cc
```diff
@@ -0,0 +1,329 @@
...
+    }
+
+    return {static_cast<unsigned char>(r), static_cast<unsigned char>(g), static_cast<unsigned char>(b), 255};
+}
+
+
+// --- Function to write content to a file and capture error messages ---
+std::pair<bool, std::vector<std::string>> writeToFileWithError(const std::string& path, const std::string& content) {
+    std::ofstream file(path, std::ios::out | std::ios::trunc);
+    std::vector<std::string> errorMessages;
```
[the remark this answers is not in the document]
calvinpark: Raylib code is ChatGPT-written soup of mess. I'm happy to apply these changes if you think it's necessary so please let me know. Thank you

system/ui/raylib/tsk-manager.cc
```diff
@@ -0,0 +1,99 @@
...
+
+    // Exit Button
+    DrawButton(exitButtonRect, kExitText, regularFont, kGray);
+    if (CheckCollisionPointRec(GetMousePosition(), exitButtonRect) && IsMouseButtonPressed(MOUSE_BUTTON_LEFT)) {
+      break;
+    }
+
+    // Extract Button
+    DrawButton(extractButtonRect, kExtractButtonText, regularFont, kGray);
+    if (CheckCollisionPointRec(GetMousePosition(), extractButtonRect) && IsMouseButtonPressed(MOUSE_BUTTON_LEFT)) {
```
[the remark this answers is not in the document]
calvinpark: The button either works or it doesn't. If it doesn't work, then there isn't much that the user can do to correct it.

tsk/manager.sh
```diff
@@ -0,0 +1,37 @@
+#!/usr/bin/env bash
+set -v
+
+CANARY="/persist/alreadyRW"
+
+cleanup() {
```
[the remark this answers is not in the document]
calvinpark: I've considered this but thought it's ultimately less effective than using a canary file

system/ui/raylib/tsk-kbd.cc
```diff
@@ -0,0 +1,329 @@
+#include "system/ui/raylib/util.h"
```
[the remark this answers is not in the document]
calvinpark: Raylib code is ChatGPT-written soup of mess. I'm happy to apply these changes if you think it's necessary so please let me know. Thank you

system/ui/raylib/tsk-kbd.cc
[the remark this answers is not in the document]
calvinpark: This key lives only on the device and is not accessible by anyone else. The device has only one user `comma` which has passwordless sudo access. Anyone with the physical ownership of the device has full control over all its files regardless of the file's ownership, and thus the umask does not provide any more security than what exists today.

system/ui/raylib/tsk-kbd.cc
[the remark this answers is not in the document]
calvinpark: Even future updates will be written by AI bots so my manual improvements will not persist. The current code complexity is low enough for the AI bots - it does not need to be lower since humans will not be reading it

tsk/manager.sh
```diff
@@ -0,0 +1,37 @@
...
+  if [ -f ${CANARY} ]; then
+    # /persist was already RW before we started.
+    # Remove the marker and don't remount as read-only.
+    rm ${CANARY}
+  else
+    # Remount /persist as read-only
+    sudo mount -o remount,ro /persist
+  fi
```
devtekve, line 14: This is smart, however, because we know `/persist` shall never be open really, I'd say keep the `cleanup` simple and always remount as `ro` even if it was `rw` prior. I am paranoid on that partition, as the risk is significant.

calvinpark: The intent was to make this PR cherrypick-able by all forks. Some forks like FP likes to write in /persist so I didn't want to disrupt that. If it's a blocker I can make a commit that changes the behavior, which some forks maintainer may choose to skip

tsk/manager.sh
```diff
@@ -0,0 +1,37 @@
...
+
+# Pip for extract.py
+TMP=/data/tmp
+mkdir TMP || true
+python3 -m pip install -r tsk/requirements.txt -t ${PYTHONPATH}
+echo "If you see this, TSK Extractor didn't run" > tsk/ext.txt
+
+# Run and exit
+cd /data/openpilot
+system/ui/tsk-manager
```
devtekve, line 37: https://github.com/sunnypilot/sunnypilot/pull/536/files#r1907735374

Basically this SH script will be the one moving the file at the end after tsk manager has run. 
The sh file would validate that the file was written, that it is valid or somerhing and only then will open the partition, write and then close it

system/ui/raylib/tsk-text.cc
```diff
@@ -0,0 +1,299 @@
...
+    Rectangle plusButtonRect = {
+        (float)GetScreenWidth() - buttonWidth - RIGHT_SIDE_MARGIN, GetScreenHeight() - 2 * buttonHeight - 2 * BUTTON_PADDING,
+        buttonWidth, buttonHeight
+    };
+    Rectangle minusButtonRect = {
+        (float)GetScreenWidth() - buttonWidth - RIGHT_SIDE_MARGIN, GetScreenHeight() - buttonHeight - BUTTON_PADDING,
+        buttonWidth, buttonHeight
+    };
+
+    // --- Main Game Loop ---
```
devtekve, line 144: haha game xD

system/ui/raylib/tsk-kbd.cc
devtekve, line 1: probably this files would be better under `....../tsk/ui/raylib` ?

selfdrive/ui/qt/offroad/settings.cc
```diff
@@ -205,6 +205,18 @@ DevicePanel::DevicePanel(SettingsWindow *parent) : ListWidget(parent) {
...
+    if (ConfirmationDialog::confirm(tr("Run Toyota Security Key Manager and reboot?"), tr("RUN"), this)) {
+      tskBtn->setEnabled(false);
+      QProcess *process = new QProcess(this);
+      // This shell script launches TSK Manager on top of OP and then reboots,
+      // so there's no need to handle hiding and restoring OP application.
+      process->start("/data/openpilot/tsk/manager.sh");
+    }
+  });
+  addItem(tskBtn);
+
```
devtekve, line 219: Add a disclaimer or warning of what's happening, I think it would be appreciated before the user continues

selfdrive/ui/qt/offroad/settings.cc
```diff
@@ -205,6 +205,18 @@ DevicePanel::DevicePanel(SettingsWindow *parent) : ListWidget(parent) {
 
   // offroad-only buttons
 
+  auto tskBtn = new ButtonControl(tr("Toyota Security Key Manager"), tr("RUN"));
+  connect(tskBtn, &ButtonControl::clicked, [this, tskBtn]() {
+    if (ConfirmationDialog::confirm(tr("Run Toyota Security Key Manager and reboot?"), tr("RUN"), this)) {
```
devtekve, line 210: also as discussed on discord, we must make sure the button is not enabled when offroad, but also make sure that even after the confirmation dialogue is closed, it will NOT run unless its still offroad

system/ui/raylib/tsk-text.cc
calvinpark: Hah, written by none other than ChatGPT

selfdrive/ui/qt/offroad/settings.cc
calvinpark: I was hoping `"Run Toyota Security Key Manager and reboot?"` conveys that message. Any suggestions?

tsk/manager.sh
devtekve: In fact, thinking about it, we don't really want to be the ones that open this partition. So for us we'd really not allow this at all, unless we absolutely must. 

Since SP will be protecting the key via sunnylink, in our case it would be more than enough to have it on the params. What we can do is try to read from /persist in hopes of another fork maybe had written it in the past, but that'd be it. And alternatively we'd allow writing it to `/data/media/0/tsk` which should survive agnos updates (but not resets afaik)

calvinpark: Okay I'll make a version that doesn't write to `/persist`. It will still try to read from it, in case a different fork allowed writing to `/persist` and archived the key there

system/ui/raylib/tsk-kbd.cc
calvinpark: Just using what's there with minimal changes. I do want to put everything inside `/data/openpilot/tsk/` but it was the fastest to keep it in `system/ui/raylib`

selfdrive/ui/qt/offroad/settings.cc
sunnyhaibin: That works. We can always update if there's a better version.

tsk/manager.sh
calvinpark: No longer applicable

PR #538, opened 2025-01-09
ui: split offroad home in its own class

sunnyhaibin:
## Summary by Sourcery

New Features:
- The offroad home screen UI has been extracted into its own class.

There is zero change to the UI.


sunnyhaibin:
@sourcery review

PR #605, opened 2025-01-22
Sentry: sets environment tag

sunnyhaibin:
## Summary by Sourcery

Enhancements:
- Use the channel type to determine the Sentry environment.

### DIFF
--- a/system/version.py
+++ b/system/version.py
@@ -10,8 +10,10 @@ from openpilot.common.basedir import BASEDIR
 from openpilot.common.swaglog import cloudlog
 from openpilot.common.git import get_commit, get_origin, get_branch, get_short_branch, get_commit_date
 
-RELEASE_BRANCHES = ['release3-staging', 'release3', 'nightly']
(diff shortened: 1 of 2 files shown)


system/version.py
```diff
@@ -110,6 +112,17 @@ def canonical(self) -> str:
...
+  @property
+  def channel_type(self) -> str:
+    if self.channel.startswith("dev-"):
+      return "development"
+    elif self.channel.startswith("staging-"):
+      return "staging"
+    elif self.tested_channel:
+      return "release"
+    else:
+      return "master"
```
devtekve, line 124:
```suggestion
      return "feature"
```

devtekve (changes requested):
Default being "master" is confusing. Set to feature

sunnyhaibin:
> Default being "master" is confusing. Set to feature

This also means we set `master` and `master-new` with the `feature` tag, which is more confusing. Can add another tag for non`master-` branches.

devtekve (approved):

PR #626, opened 2025-01-30
Support for brand/make/model

Discountchubbs:
***** Refactor *****

Added support in-vehicle panel for toggles targeting brand (parent group e.g., 'hyundai'), make (e.g., Kia), and model (e.g., Ioniq 5).

Offroad to onroad transition. Enables restricting toggles (such as tuning toggles) to only be enabled/disabled when the vehicle is in offroad status, preventing potential conflicts.

Added reboot prompts for toggles that require a reboot if enabled while on the road. 

Added the ability to restrict toggles "enable" status to require openpilot longitudinal toggle to be in the "on" state.

I also added an error log catcher to the software settings for potential implementation in the dev branch for user access.

I included many examples in the vehicle_panel.cc and vehicle_panel.h files for reference (denoted in "//" so the boot/compile sequence remains unaffected).

I left the examples to reference the various ways we can use the brand, make, or model to classify toggles for users.

## Summary by Sourcery

New Features:
- Added support for toggles based on vehicle brand, make, and model.


selfdrive/ui/qt/offroad/software_settings.cc
```diff
@@ -83,6 +83,14 @@ SoftwarePanel::SoftwarePanel(QWidget* parent) : ListWidget(parent) {
...
   addItem(uninstallBtn);
 
+  // error log button
+  auto errorLogBtn = new ButtonControl(tr("Error Log"), tr("VIEW"), tr("View the error log for openpilot crashes."));
+  connect(errorLogBtn, &ButtonControl::clicked, [=]() {
+    std::string txt = util::read_file("/data/community/crashes/error.log");
+    ConfirmationDialog::rich(QString::fromStdString(txt), this);
+  });
+  addItem(errorLogBtn);
+
```
sunnyhaibin, line 93: Now that we have the Developer panel, let's open a new PR and have it under there!

selfdrive/ui/sunnypilot/qt/offroad/settings/vehicle_panel.cc
```diff
@@ -20,11 +20,101 @@ VehiclePanel::VehiclePanel(QWidget *parent) : QFrame(parent) {
...
+
+  QObject::connect(uiState(), &UIState::offroadTransition, [=](bool offroad) {
+    isOnroad = !offroad;
+    updateCarToggles();
+  });
+
   ScrollViewSP *scroller = new ScrollViewSP(list, this);
   vlayout->addWidget(scroller);
 
   QObject::connect(uiState(), &UIState::offroadTransition, platformSelector, &PlatformSelector::refresh);
```
sunnyhaibin, line 33: L33 already connects `UIState::offroadTransition` to `PlatformSelector::refresh`, so we can consolidate the 2 `QObject::connect` together.

Let's keep using `offroad` instead of `isOnroad`, so it's more consistent with what we already use for `DeveloperPanel` and `PlatformSelector`.

selfdrive/ui/sunnypilot/qt/offroad/settings/vehicle_panel.cc
```diff
@@ -20,11 +20,101 @@ VehiclePanel::VehiclePanel(QWidget *parent) : QFrame(parent) {
...
 
+  QObject::connect(platformSelector, &PlatformSelector::clicked, [=]() {
+    QTimer::singleShot(100, this, &VehiclePanel::updateCarToggles);
+  });
+
+  QTimer *timer = new QTimer(this);
+  QObject::connect(timer, &QTimer::timeout, this, &VehiclePanel::updateCarToggles);
+  timer->start(1000);
```
sunnyhaibin, line 41: The `QTimer` used here along with the `platformSelector` button currently delays for 100 ms after being clicked, then invokes `updateCarToggles` every 1 second. Do we have to refresh the panel continuously, even with no user interaction?

selfdrive/ui/sunnypilot/qt/offroad/settings/vehicle_panel.cc
```diff
@@ -20,11 +20,101 @@ VehiclePanel::VehiclePanel(QWidget *parent) : QFrame(parent) {
...
+void VehiclePanel::showRebootPrompt() {
+  QString question = tr("Changing this setting requires a reboot to take effect.\n\n"
+                       "Reboot now or later?");
+
+  reboot_confirmed = ConfirmationDialog::confirm(question, tr("Reboot Now"), this);
+  if (reboot_confirmed && !reboot_completed) {
+    params.putBool("DoReboot", true);
+    reboot_completed = true;
+  }
+}
```
sunnyhaibin, line 56: For most options and toggle changes, we usually shouldn't need to perform a full reboot to take effect. How about if we invoke "Always Offroad" mode we have in sunnypilot? We introduced "Offroad Mode" in sunnypilot a little while back. This practically allows the settings to take effect without a full reboot:
- https://github.com/sunnypilot/sunnypilot/pull/596

selfdrive/ui/sunnypilot/qt/offroad/settings/vehicle_panel.cc
```diff
@@ -20,11 +20,101 @@ VehiclePanel::VehiclePanel(QWidget *parent) : QFrame(parent) {
...
+    case ToggleState::DISABLED_LONGITUDINAL:
+      toggle->setEnabled(false);
+      toggle->setToolTip(LONGITUDINAL_MSG);
+      break;
+    case ToggleState::DISABLED_DRIVING:
+      toggle->setEnabled(false);
+      toggle->setToolTip(DRIVING_MSG);
+      break;
+  }
+}
```
sunnyhaibin, line 88: I like this! 😄

selfdrive/ui/sunnypilot/qt/offroad/settings/vehicle_panel.cc
```diff
@@ -20,11 +20,101 @@ VehiclePanel::VehiclePanel(QWidget *parent) : QFrame(parent) {
...
+void VehiclePanel::updateCarToggles() {
+  //bool hasOpenpilotLong = params.getBool("ExperimentalLongitudinalEnabled");
+
+  QString platform_name = QString::fromStdString(params.get("CarPlatformName"));
+  //if (platform_name.isEmpty()) {
+  //  hkgtuningToggle->setVisible(false);
+   // return;
+  //} To hide the toggle if no platform is selected
+
+  QVariantMap platform_data = platformSelector->loadPlatformList()[platform_name];
```
sunnyhaibin, line 99: We should avoid calling this too often. I will expose another param to cache this field persistently in the other PR, so we don't need to invoke whenever user tries to interact with the toggles.

sunnyhaibin:
For this PR, I propose getting `getToggleState` and `updateToggleState` cleaned up and merged, and we can have the rest open in other PRs to have them merged in small bits. cc @devtekve

selfdrive/ui/sunnypilot/qt/offroad/settings/vehicle_panel.cc
devtekve: Me too!!!!

Discountchubbs:
**Replaced setToolTip with description notifications (like how notifications are implemented in sunnylink).
**Added cachedPlatformData, which caches platform data from getPlatformBundle once and saves it for use when the user calls (it updates the cache when the user removes and adds a vehicle, so toggles don't stay visible).
**Removed the 1-second refresh rate and only refresh on user action.
**Removed isOnroad and instead used offroad for toggle handling
**Removed redundant QObject::Connect
**Removed examples





20250131_223207000_iOS
20250131_223150000_iOS

sunnyhaibin:
@Discountchubbs can you rebase?

PR #650, opened 2025-03-08
CI: use one label for `runs-on`

devtekve:
use one label for runs-on instead of multi-label and disable process-replay


devtekve:
@sourcery-ai update title and description

sunnyhaibin (approved):

PR #659, opened 2025-03-14
Car interface: custom safety params

sunnyhaibin:
## Summary by Sourcery

This pull request introduces custom safety parameters for the car interface, specifically for SunnyPilot. It adds a new field to CarParamsSP for custom safety flags and updates the Panda code to handle these new parameters.

- https://github.com/sunnypilot/opendbc/pull/81
- https://github.com/sunnypilot/panda/pull/75

## Summary by Sourcery

Introduces custom safety parameters for the car interface, specifically for SunnyPilot, by adding a new field to CarParamsSP for custom safety flags and updating the Panda code to handle these new parameters.

Enhancements:
- Adds a new field to CarParamsSP for custom safety flags.
- Updates the Panda code to handle the new safety parameters.


selfdrive/pandad/panda.cc
```diff
@@ -62,8 +62,8 @@ void Panda::set_safety_model(cereal::CarParams::SafetyModel safety_model, uint16
   handle->control_write(0xdc, (uint16_t)safety_model, safety_param);
 }
 
-void Panda::set_alternative_experience(uint16_t alternative_experience) {
-  handle->control_write(0xdf, alternative_experience, 0);
```
devtekve, line 66: We went from sending 0 to sending something. I am guessing the 0 was params from stock?

selfdrive/pandad/panda_safety.cc
```diff
@@ -56,16 +57,20 @@ std::string PandaSafety::fetchCarParams() {
   if (!params_.getBool("ControlsReady")) {
     return {};
   }
-  return params_.get("CarParams");
+  return {params_.get("CarParams"), params_.get("CarParamsSP")};
```
devtekve, line 60: I think I would feel better if this was a struct and not an array, this ain't python xD

selfdrive/pandad/pandad.h
```diff
@@ -16,8 +16,8 @@ class PandaSafety {
 
 private:
   void updateMultiplexingMode();
-  std::string fetchCarParams();
-  void setSafetyMode(const std::string &params_string);
+  std::vector<std::string> fetchCarParams();
```
devtekve, line 19: Maybe struct instead of vector/array

sunnypilot/mads/helpers.py
```diff
@@ -55,7 +54,7 @@ def set_car_specific_params(self, CP, CP_SP):
       hyundai_cruise_main_toggleable = True
       if hyundai_cruise_main_toggleable:
         CP_SP.flags |= HyundaiFlagsSP.LONGITUDINAL_MAIN_CRUISE_TOGGLEABLE.value
-        CP.safetyConfigs[-1].safetyParam |= HyundaiSafetyFlags.FLAG_HYUNDAI_LONG_MAIN_CRUISE_TOGGLEABLE.value
+        CP_SP.safetyParam |= HyundaiSafetyFlagsSP.LONG_MAIN_CRUISE_TOGGLEABLE
```
devtekve, line 57:
```suggestion
        CP_SP.safetyParam |= HyundaiSafetyFlagsSP.LONG_MAIN_CRUISE_TOGGLEABLE.value
```
I am pretty sure we need to use `.value`, dont we?

selfdrive/pandad/panda.cc
sunnyhaibin: It used to send 0 because stock doesn't use this field with this endpoint, so it is freed up for us to send whatever we want via SPI/USB.

sunnypilot/mads/helpers.py
sunnyhaibin: For `HyundaiSafetyFlagsSP`, since we defined them directly as integer constants (not as an Enum), we don't need the `.value`. The constant `LONG_MAIN_CRUISE_TOGGLEABLE` is already the integer value that can be used directly in bitwise.

devtekve (approved):

PR #676, opened 2025-03-18
Refactor - SLC: Simplify Dynamic Personalty Logic

USMortality:
**Description**

This change simplifies the logic of the dynamic personality feature. It uses the already defined standard offsets of T_FOLLOW, but adjusts them gradually from 75% to 100% from 0 to 36 speed. I think this simplifies the logic, and avoids hard coded values.

I'm not entirely sure why Kumar chose 36.01, as I could not find a PR, but for now this mirrors his logic.

**Verification**

Manually tested in Kia Sportage 2023 + Comma C3.


devtekve:
@USMortality This will wait a little bit until we bring SLC back on the rewrite to target `master-new`

USMortality:
Hi, would be nice to know why this was closed. Thanks :)

PR #677, opened 2025-03-18
Feature - SLC: Override Speed Limit Offset via Gas Pedal.

USMortality:
**Description**

This change adds a new subtle feature, which lets you override the SLC, offset via gas pedal. Say you have set cruise max to 60, and you are in a 25 zone, currently it would use +3 (+10%), now the driver can simply increase this via gas pedal, say to 35 and the speed limit sign will show 25 +10.
This is reset, on next speed limit change or when user adjusts the set cruise speed.

**Verification**

Manually tested in Kia Sportage 2023 + Comma C3.

## Summary by Sourcery

Enhancements:
- Add gas pedal override for speed limit offset.


devtekve:
@USMortality This will wait a little bit until we bring SLC back on the rewrite to target `master-new`

USMortality:
Hi, would be nice to know why this was closed. Thanks :)

PR #704, opened 2025-03-23
NNLC: bump max similarity for higher accuracy

royjr:
## Summary by Sourcery

Increase the minimum similarity score for selecting a neural network model from 0.8 to 0.9. This change ensures that only models with a higher degree of similarity to the vehicle's fingerprint are used, potentially improving the accuracy and reliability of the neural network-based longitudinal control.


sunnyhaibin (approved):

sunnyhaibin (approved):
LGTM

PR #714, opened 2025-03-25
modeld: Enforce float type for `lag_adjusted_curvature`

devtekve:
Found that when we try to fill_model_msg and the model output contains `lat_planner_solution` (BDv2 does) then the serialization of capnp fails beause the value returned by `get_lag_adjusted_curvature` is now `numpy.float64` instead of `float` as it used to be due to the removal of numpy fast and subsequent update on https://github.com/sunnypilot/sunnypilot/commit/141a7f5f72d6ed1c5b466a30ee55f5b58f3dac46 

image

This will make sure the value returned is converted to python's primitive float

## Summary by Sourcery

Bug Fixes:
- Ensures that the value returned by `get_lag_adjusted_curvature` is converted to a Python float to avoid serialization errors.


Discountchubbs (approved):
cc @sunnypilot/dev-internal 

This modification to return a `float` should resolve the numpy.float64 error encountered in the file `drive_helpers.py` when support for `numpy_fast` was deprecated.

PR #720, opened 2025-03-26
Events: Encapsulate events in OnroadEventSP with a list

devtekve:
The `onroadEventsSP` message now encapsulates a list of `Event` objects, improving flexibility and organization. Adjustments were made in the Cap'n Proto definitions and Python messaging code to support this new structure. This change enhances the handling of on-road events by grouping them more effectively.

## Summary by Sourcery

Refactor the OnroadEventSP structure to encapsulate events in a list, improving event handling and organization in the messaging system

Enhancements:
- Modify Cap'n Proto definitions to support a more flexible event structure by introducing a list of Event objects within OnroadEventSP

Chores:
- Update messaging code to work with the new event list structure in OnroadEventSP

### DIFF
--- a/cereal/custom.capnp
+++ b/cereal/custom.capnp
@@ -102,37 +102,41 @@ struct LongitudinalPlanSP @0xf35cc4560bbf6ec2 {
 }
 
 struct OnroadEventSP @0xda96579883444c35 {
-  name @0 :EventName;
+    events @0 :List(Event);
 
-  # event types
-  enable @1 :Bool;
-  noEntry @2 :Bool;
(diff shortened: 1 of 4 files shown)


sunnyhaibin:
- [ ] test to create a file (bytes) with empty services, assert any failure when trying to read it with the upstream cereal

devtekve:
> * [ ]  test to create a file (bytes) with empty services, assert any failure when trying to read it with the upstream cereal

This PR does not intend to include any test to prevent a mistake like that happening in the future, this is a bug fix to prevent the routes to keep getting created wrong. The test can be implemented later

devtekve:
Do not merge this

PR #727, opened 2025-03-29
Revert "H7: detect board type with second ID pin group"

sunnyhaibin:


devtekve (approved):

devtekve:
Validate staging too just in case

PR #729, opened 2025-03-30
Revert "Revert "H7: detect board type with second ID pin group""

sunnyhaibin:
Reverts sunnypilot/sunnypilot#727


sunnyhaibin:
This was causing controls to exit with some Hyundai. Need to investigate and rebuild.

PR #784, opened 2025-04-04
MADS: Soft disable for non-Drive forward gear if in motion

sunnyhaibin:
## Summary by Sourcery

Bug Fixes:
- Add a condition to only trigger soft disable for wrong gear when the vehicle is not moving (at standstill)


wtogami:
> non-Drive forward gear

This wording is confusing. Perhaps "downshifted gear"?

sunnyhaibin:
> > non-Drive forward gear
> 
> This wording is confusing. Perhaps "downshifted gear"?

“Downshifted gear” doesn’t quite capture the full scope here. This change applies to other non-Drive forward gears like Manumatic (M), Sport (S), and Brake (B) modes, not just traditional downshifts.

PR #787, opened 2025-04-04
CI: bump UI Report timeout

sunnyhaibin:
## Summary by Sourcery

CI:
- Adjusted the timeout minutes for the Create Test Report step in the CI workflow, increasing the timeout when the frames cache is not hit

### DIFF
--- a/.github/workflows/selfdrive_tests.yaml
+++ b/.github/workflows/selfdrive_tests.yaml
@@ -332,13 +332,21 @@ jobs:
         with:
           submodules: true
       - uses: ./.github/workflows/setup-with-retry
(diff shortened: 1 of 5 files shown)


sunnyhaibin:
If "Create UI Report" fails in the future due to assert during the frames_cache step in any PR:
1. Comment this line: https://github.com/sunnypilot/sunnypilot/blob/84b934033c7c11d89874ab059897d8d5a352cf56/tools/lib/framereader.py#L493
2. Push and let it build and save the cache to GH Actions.
3. Uncomment the line from step 1.
4. Push and confirm that it is using the newly created cache.

PR #803, opened 2025-04-10
Add Reverse ACC Change parameter and update cruise control logic

devtekve:

### DIFF
--- a/selfdrive/car/card.py
+++ b/selfdrive/car/card.py
@@ -131,6 +131,7 @@ class Car:
 
     # Dynamic Experimental Control
     self.dynamic_experimental_control = self.params.get_bool("DynamicExperimentalControl")
+    self.reverse_acc_change = self.params.get_bool("ReverseAccChange")
 
     openpilot_enabled_toggle = self.params.get_bool("OpenpilotEnabledToggle")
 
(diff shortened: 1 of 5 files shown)


selfdrive/ui/qt/offroad/developer_panel.cc
```diff
@@ -30,6 +30,12 @@ DeveloperPanel::DeveloperPanel(SettingsWindow *parent) : ListWidget(parent) {
   });
   addItem(longManeuverToggle);
 
+  reverseAccChangeToggle = new ParamControl("ReverseAccChange", tr("Reverse ACC Change"), tr("Enable this to change swap the ACC long press and short press behavior."), "");
```
Discountchubbs, line 33: ACC Reverse Short Press Speed Change

selfdrive/car/card.py
```diff
@@ -131,6 +131,7 @@ def __init__(self, CI=None, RI=None) -> None:
 
     # Dynamic Experimental Control
     self.dynamic_experimental_control = self.params.get_bool("DynamicExperimentalControl")
+    self.reverse_acc_change = self.params.get_bool("ReverseAccChange")
```
Discountchubbs, line 134: `ReverseSpeedPress` Or something like that, so it doesn't seem like we are rocking OP in reverse gear.

Discountchubbs (changes requested):
Quick naming changes before PR merge, and focus on Toyota, Rivian, Tesla in a different PR.

PR #810, opened 2025-04-12
ci: Remove redundant Panda build step from prebuilt workflow

sunnyhaibin:
The Panda build step was unnecessary as it is not utilized in this workflow. This change simplifies the workflow and reduces redundant actions, improving efficiency.

## Summary by Sourcery

CI:
- Removed redundant Panda build step from the GitHub Actions workflow to simplify and improve workflow efficiency


devtekve (commented):
I am reviewing that the prebuilt workflow still works with this https://github.com/sunnypilot/sunnypilot/actions/runs/14417945933 and if so I will merge

devtekve (approved):

PR #812, opened 2025-04-12
update urls untill sunnypilot fork installer is ready

Stupefacient:
## Summary by Sourcery

Documentation:
- Update installation URLs for staging and development branches to use temporary hosting domain


devtekve:
@Stupefacient the installer _**is**_ ready though, we don't really need this

PR #822, opened 2025-04-15
NNLC: decreased low-speed factor

twilsonco:
The NNLC reimplementation was missing a few things from the old SP implementation. One of them was a [decreased low-speed factor for NNLC](https://github.com/sunnypilot/sunnypilot/blob/1820d16e96c9ad2ee7ad980e0d6a557c2fdfae39/selfdrive/controls/lib/latcontrol_torque.py#L29). 

The way I addressed that here results in redundantly calculating the setpoint and measurement, but it's a small amount to recalculate, and this way it doesn't touch the torque controller code, which is a priority of the NNLC reimplementation in the `-new` branches.

## Summary by Sourcery

Implement a low-speed factor adjustment for Neural Network Lateral Control (NNLC) to improve lateral control at low speeds

New Features:
- Introduce low-speed scaling mechanism using interpolation between predefined speed points

Enhancements:
- Add a low-speed interpolation factor to adjust lateral acceleration and curvature calculations in the NNLC implementation


sunnyhaibin (approved):
LGTM. Thanks for catching this!

PR #823, opened 2025-04-15
NNLC: fix weak torque in high lat-accel turns

twilsonco:
The NNLC reimplementation was missing a few things from the old SP implementation. One of them was ["error blending"](https://github.com/sunnypilot/sunnypilot/blob/1820d16e96c9ad2ee7ad980e0d6a557c2fdfae39/selfdrive/controls/lib/latcontrol_torque.py#L240), which was a workaround for weak error response in high magnitude lat accel curves for cars whose NNLC model was trained with insufficient high magnitude lat accel data. 

This reintroduces error blending, which, under high lat accel conditions, results in an extra evaluation of the NNLC model.

## Summary by Sourcery

Improve Neural Network Lateral Control (NNLC) error response for high lateral acceleration scenarios by introducing error blending

Bug Fixes:
- Address weak error response in high lateral acceleration curves by dynamically adjusting the error calculation

Enhancements:
- Implement an error blending technique to strengthen the torque response for NNLC models with insufficient high-magnitude lateral acceleration training data


wtogami:
I noticed a MAJOR improvement. Previous NNLC would understeer and fail medium to high speed curves. Now with this PR it handled those curves with significantly less need of human intervention.

cb6aec3054a67f94/000000a0--9b8a0398d7

sunnyhaibin (approved):
LGTM with the sanity check test. Thanks for the fix!

PR #827, opened 2025-04-15
ui: display recorded timestamp on error view

Discountchubbs:
> [!NOTE]
> Depends on:
> * #1051

Add a timestamp in bold on error log representing last modified date and time.

Screenshot 2025-07-16 at 12 42 32 PM

## Summary by Sourcery

Show the last modified timestamp in bold at the top of the error log dialog

Enhancements:
- Include the file’s last modified date and time in the error log view
- Add QFileInfo to read and format the timestamp before displaying the log


Discountchubbs:
image
image

devtekve (approved):

devtekve (changes requested):
Let me put a little hold on this, maybe clearing the error log at boot might be too soon..

devtekve:
@Discountchubbs instead of clearing the file can we actually add a timestamp on the top? @sunnyhaibin had something like that in the past i believe. It would help more than clearing it. Maybe we should find another way to clear it ? I am worried that users might reboot and that's it, error log is gone. (a user had it happen on `dev-c3-new`

PR #832, opened 2025-04-17
Longitudinal: Custom ACC Speed Increments

nayan8teen:
Reverse the "Reverse ACC" change ;)
And allow custom values for short & long acc button press

### DIFF
--- a/selfdrive/car/card.py
+++ b/selfdrive/car/card.py
@@ -131,7 +131,6 @@ class Car:
 
     # Dynamic Experimental Control
     self.dynamic_experimental_control = self.params.get_bool("DynamicExperimentalControl")
-    self.reverse_acc_change = self.params.get_bool("ReverseAccChange")
 
     openpilot_enabled_toggle = self.params.get_bool("OpenpilotEnabledToggle")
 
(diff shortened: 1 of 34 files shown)


selfdrive/car/card.py
```diff
@@ -219,7 +229,11 @@ def state_update(self) -> tuple[car.CarState, structs.RadarDataT | None]:
     if can_rcv_valid and REPLAY:
       self.can_log_mono_time = messaging.log_from_bytes(can_strs[0]).logMonoTime
 
-    self.v_cruise_helper.update_v_cruise(CS, self.sm['carControl'].enabled, self.is_metric)
+    custom_acc_increment = None
+    if self.custom_acc_increments_enabled:
+      custom_acc_increment = {"short_press": int(self.custom_acc_short_increment), "long_press": int(self.custom_acc_long_increment)}
```
devtekve, line 234: While this is working, I am not entirely sold on it. It feels weird to get the params from card just to pass them to cruise.py (feels like prop drilling from react). Furthermore, having it as dict instead of an object is itching on my head :D 

And finally, I kind of have the feeling maybe this should be extracted on a module under `/sunnypilot/...` folder like a utils or something. 

So Food for thought is: Maybe cereal would be a better way for this? it would give us an object, it would allow us to see it easier on the logs and it would decouple it from "params" itself (you'd still load from the params, but like runtime wise it would be cereal)

PR #834, opened 2025-04-19
UI : Visual indicator for blind spot

rav4kumar:


sunnyhaibin (commented):
- [x] Split `model.cc` and `model.h` into its own files for custom UI

devtekve (approved):

devtekve (approved):

PR #846, opened 2025-04-22
Update SSH install method git clone script. Add -b

Paperboypaddy:
Fix clone script adding -b for the branch name.


sunnyhaibin:
Thanks for the PR. We're refactoring this branch with another approach, so this branch will no longer be maintained and target to merge into `master`. If you see anything on the new docs that needs to be updated, please feel free to open another PR to fix!

PR #850, opened 2025-04-23
neural_network_data: TOYOTA_RAV4_PRIME NNLC tuning gen 1

wtogami:
Trained on 2023 RAV4 Prime XSE running fuzzy TOYOTA_RAV4_TSS2_2022.json so you could consider this to be a more refined almost gen 2 tuning.

image

twilsonco's analysis: "But the eps torque model looks really good! Yours looks nearly identical [to TOYOTA_RAV4_TSS2_2022]. It has a subtler instantaneous lateral jerk response. More gradual instantaneous jerk response."

This confirms the new tuning despite generated independent inputs is close to its predecessor.

Drive Test results
* It feels more confident and in control with more torque when appropriate.
* Seems to be able to handle several high torque curves that OP lat or NNLC was previously failing.

Testing Wishlist
* This was trained and tested on the 2023 RAV4 Prime XSE with 19 inch wheels. It should be tested by the SE trim with 18 inch wheels.


wtogami:
By kumar's request pushed again. No change except `neural_network_data` is now pointing at the submodule `master` instead of a PR.

rav4kumar (approved):

Issue #828, opened 2025-04-15
Allow starting in always offroad mode


devtekve:

wtogami:
The other reason is various EV modes like Toyota "Room Mode" is seen by OP as "Ignition On" but the car can't move. I use that as an air conditioned office sometimes. I don't want the camera to record me and my family during those times, and the route recorded during that time is useless.

Can a button to turn off Offroad mode be added to the main screen? Then you don't need to open the settings and scroll down to turn SP back on.

nayan8teen:
Is this already being worked on? If not, I could take a stab at it.

devtekve:
> [@devtekve](https://github.com/devtekve) Is this already being worked on? If not, I could take a stab at it.

Go for it ! We haven’t been working on it yet.

I had in mind 2 things: properly detecting cars that have that sort of “camping” mode, and maybe having the toggle to start always forced off-road and when that’s active, show on the home page a button to go onroad (similar to where the firehouse button is or where the navigation destination used to be).

I would not add anything in the onroad screen, so no button to go off-road from the onroad page. You should only ever be able to go off-road from the settings and when the control is not active

wtogami:
Idea 4. Route recording geofencing

Define an area. If you are within that area it stops recording.*
The goal here is privacy to be able to share routes publicly without showing exactly where you live.

* Instead of stop recording entirely, what if it is recorded to a separate local store that isn't uploaded to Connect? That way if you need that portion of the recording it is available in a different way.

nayan8teen:
Started playing with this. When we say "start in offroad mode", should that be only at boot? Or should we force always offroad mode every time the device transitions from onroad->offroad so that it will NEVER go into onroad state without the user explicitly exiting always offroad.

wtogami:
"start in offroad mode" at boot.

That would be the simplest workaround for the above variety of other problems.

nayan8teen:
> "start in offroad mode" at boot.
> 
> That would be the simplest workaround for the above variety of other problems.
> 
> 

Yeah, that's what I started with. Just needs some cleanup before I submit a PR for it - probably later tonight.

nayan8teen:
PR Created: https://github.com/sunnypilot/sunnypilot/pull/861

nayan8teen:
For adding an Exit button on Offroad screen.. preference between these 2?

![Image](https://github.com/user-attachments/assets/2a384f1e-c897-43b7-8e2b-6d744e8145b9)

[...]

PR #861, opened 2025-04-29
Device: Start in Always Offroad

nayan8teen:
1. Resolves [Allow starting in always offroad mode #828](https://github.com/sunnypilot/sunnypilot/issues/828)
2. Re-arranged Offroad related controls on Device screen
3. Created an Inline layout for ButtonParamControlSP similar to OptionControlSP


wtogami:
I didn't try this yet.

Just adding a comment: I would like it if the turning off Offroad button was closer to the top of the menu so I don't need to scroll down to find it.

nayan8teen:
> I didn't try this yet.
> 
> Just adding a comment: I would like it if the turning off Offroad button was closer to the top of the menu so I don't need to scroll down to find it.

I'm thinking of making an exit offroad button available on the offroad home screen instead to make it easier. That'll be a separate PR though as it doesn't block any function of this one.

nayan8teen:
image
Don't know what's going on with the preview. Didn't touch anything around driver camera at all.

devtekve:
@nayan8teen 
image

could we just make it a toggle instead like on the other places? 
image


And can we them back above thee other buttons as they were ? I think the separation is a lot and breaks the cohesion

nayan8teen:
> @nayan8teen image
> 
> could we just make it a toggle instead like on the other places? image
> 
> And can we them back above thee other buttons as they were ? I think the separation is a lot and breaks the cohesion

That's what I had initially, but it looked very out-of-place because of the different alignments:

image

devtekve:
> > @nayan8teen image
> > could we just make it a toggle instead like on the other places? image
> > And can we them back above thee other buttons as they were ? I think the separation is a lot and breaks the cohesion
> 
> That's what I had initially, but it looked very out-of-place because of the different alignments:
> 
> image

What if we make a subpanel for the offroad behavior? So it'd be a button like we have to customize other stuff... Like under steering there's the MADS settings

devtekve:
Or maybe we can add a panel on the left that's called "Privacy" and maybe there we will add other stuff besides the start offroad in the future (for now just the offroad) 

but maybe really there's people that want to collect zero logs of anything so i we add a feature like that, that'd be the panel for it

wtogami:
It is confusing to be under Privacy. Reasons to want boot into Offroad include privacy and non privacy reasons. I want to avoid recording useless non moving routes. I also want to control when Onroad begins so it calibrates only on painted lanes further away from my house.

wtogami:
Tested it. It does exactly as advertised. But I realized there is a point of confusion.

If the device has powered off, this works exactly as intended.

But I was confused after I started my car and it went directly into Onroad mode when it saw ignition. It turns out the device was not powered off just sleeping with screen off. To achieve the expected behavior it may need to revert back to Offroad mode when it goes to sleep?

nayan8teen:
> Tested it. It does exactly as advertised. But I realized there is a point of confusion.
> 
> If the device has powered off, this works exactly as intended.
> 
> But I was confused after I started my car and it went directly into Onroad mode when it saw ignition. It turns out the device was not powered off just sleeping with screen off. To achieve the expected behavior it may need to revert back to Offroad mode when it goes to sleep?

That is the current expected behavior - hence the name "Device Boot Mode". Forcing off-road after interactivity timeout (aka sleep) can be done, but could also lead to user confusion.

wtogami:
> That is the current expected behavior - hence the name "Device Boot Mode". Forcing off-road after interactivity timeout (aka sleep) can be done, but could also lead to user confusion.

I think it's confusing as is because the user doesn't generally control the difference between sleep and powered off.

Perhaps this needs a different name. If you enable whatever this is, you don't expect routes to immediately begin recording when you turn the car on.

wtogami:
**Device: Start in Always Offroad**

The title here sounds like the intended effect.
The "Boot" part of the UI happens less often than wake from suspend. What better name to give it? I don't know.

nayan8teen:
@wtogami can you test the latest push? Haven't changed the description yet, but the offroad behavior should now apply to sleep as well.

wtogami:
Your branch requires conflict fix for recently merged things into `master-new`

wtogami:
`start-offroad` exposed a new problem. One of the reasons I wanted this is to be able to change settings after starting the car before I move without recording useless 5 second routes. I especially want to change settings immediately after swapping the device into the other car. But starting Offroad actually breaks your ability to change certain settings after car swap because it doesn't become aware of the new fingerprint until after Offroad is turned off. So it requires going Onroad then Offroad again. It would be perfect if it can read the fingerprint without going Onroad.

selfdrive/ui/sunnypilot/qt/widgets/controls.h
```diff
@@ -249,14 +249,11 @@ class ButtonParamControlSP : public AbstractControlSP_SELECTOR {
...
       button_param_layout->setSpacing(0);
+      spacingItem = nullptr;
       if (!title.isEmpty()) {
         main_layout->removeWidget(title_label);
         hlayout->addWidget(title_label, 1);
       }
-      if (spacingItem != nullptr && main_layout->indexOf(spacingItem) != -1) {
-        main_layout->removeItem(spacingItem);
-        spacingItem = nullptr;
-      }
```
wtogami, line 259: Is this intentional? Why did they have this?

wtogami:
This works great. The rare car swapping use case will be filed as separate issue.

I wrote a question in the diffs. Weird I don't see it in the PR history.

wtogami:
The combination of #861 and #872 together seem to work exactly as intended.

selfdrive/ui/sunnypilot/qt/offroad/settings/device_panel.cc
```diff
@@ -80,6 +80,21 @@ DevicePanelSP::DevicePanelSP(SettingsWindowSP *parent) : DevicePanel(parent) {
...
   addItem(maxTimeOffroad);
 
+  toggleDeviceBootMode = new ButtonParamControlSP(
+    "DeviceBootMode",
+    tr("Wake-Up Behavior"),
+    "",
+    "",
+    {"Default", "Offroad"},
+    375,
+    true);
```
devtekve:
```suggestion
    toggleDeviceBootMode = new ButtonParamControlSP("DeviceBootMode", tr("Wake-Up Behavior"), "", "", {"Default", "Offroad"}, 375, true);
```

devtekve (approved):

PR #881, opened 2025-05-05
bug: fix braking issue while experimental and using old model

nayan8teen:
## Summary by Sourcery

Fix longitudinal planner behavior for non-stock models by adjusting acceleration target and stop conditions

Bug Fixes:
- Ensure correct acceleration target selection for different model runners

Enhancements:
- Add support for handling non-stock model runners in longitudinal planning


Discountchubbs (approved):
@devtekve

devtekve (approved):
Let's roll with this or we will be having other users that can't load models as they've been removed from the selector due to this bug to prevent safety issues

PR #908, opened 2025-05-11
ui: display actual fingerprint name with auto-fingerprint

nayan8teen:
keep brand from platform package

## Summary by Sourcery

Enhancements:
- Preserve the original platform name when displaying fingerprint information in the UI.


sunnyhaibin (approved):
Confirmed the settings are visible with auto-fingerprint.

PR #917, opened 2025-05-11
Update controlsd.py

SashaLikh:


devtekve:
@SashaLikh why not make this PR to upstream though? https://github.com/commaai/openpilot/blob/master/selfdrive/controls/controlsd.py#L94

PR #925, opened 2025-05-16
NNLC: `RAV4_PRIME` tuning Gen 2

wtogami:
NNLC Gen 2 model for 2021-2023 RAV4 Prime.

**Before: `TOYOTA_RAV4_PRIME.json` Gen 1** was merged in #850 on April 23rd, 2025.

RAV4PRIME lat_accel_vs_torque

**After: `TOYOTA_RAV4_PRIME.json` Gen 2 May 15th** trained from logs running Gen 1 and minimizing touching the steering wheel. The Gen 1 model made steering far more accurate. Minimizing touching the steering wheel substantially reduced the measured **Driver torque** in the chart below.

RAV4PRIME lat_accel_vs_torque

**Before: Gen 1**

TOYOTA_RAV4_PRIME_torque_adjusted_eps-a

**After: Gen 2 May 15th**

TOYOTA_RAV4_PRIME_torque_adjusted_eps-a

**Drive Test results**
* Extensive tests indicate improved lat precision at low, medium, and high speeds.
* It handled some tricky curves 51-75mph that Gen 1 could not.
* No regressions reported.

**Vehicles Tested**
* 2023 RAV4 Prime XSE with 19in wheels
* 2021 RAV4 Prime SE with 18in wheels

**Test Plan**
* Regressions? Does it handle anything worse than Gen 1?
* Please report anything you notice different from Gen 1.
 * Describe if anything changed. Does anything feel different? Did it improve anything?

## Summary by Sourcery

Introduce draft Gen2 NNLC model for 2021–2023 Toyota RAV4 Prime to improve lateral control and reduce driver torque.

New Features:
- Add May 15th Gen2 NNLC configuration file for RAV4 Prime, replacing the Gen 1 model.

Enhancements:
- Improve lateral precision at low, medium, and high speeds.
- Reduce measured driver torque by minimizing steering corrections.
- Enable handling of high-speed curves (51–75 mph) that Gen 1 struggled with.

Documentation:
- Update performance charts comparing lateral acceleration vs torque for Gen 1 and Gen 2 models.


wtogami:
This has noticeable improvements at all speeds without reported regressions.

That said there will be additional training runs with more data to see if it can be refined further before merge.

wtogami:
> [!NOTE]
> **_devtekve's edit_:** On the response, when twil refers to columns, he's referring to the original full screenshot, the screenshots attached are already the cropped comparison for ease of readability

@twilsonco review:

image
> The difference are nuanced, as they typically are when it comes to changes between iterations of NNLC model generations. 
> First, in the left column (instantaneous lateral jerk response), see how the new learned response is more gradual. before, the ±0.25 lines are almost the same as the ±1.0 lines. Now the lines are more evenly spaced. This means the increase in steer torque in response to lateral jerk will be more gradual, so when it wants to change the steering angle slowly vs quickly, it will get a different amount of additional steer torque to achieve that steering wheel movement. This should make it less jittery when making smaller corrections, since those won't provoke as much of a torque response.

> Also the directional dependence of instantaneous lateral jerk is more apparent here. This only comes into play when you're at high enough lateral acceleration, and corresponds to the fact that, when you're already in a curve (e.g. to the right), it takes a different amount of steer torque to initiate steering wheel movement into the curve (steering more to the right) vs back to center (to the left). In the plots, you can see that the learned response is to decrease toque more to return to center, compared to increasing it to tighten the curve. 

> That was also the case with the "before" plots, but now again it's more gradual. So when you're in a curve and it just starts turning back to center, it should do so in smaller, less noticable steps of steer angle change.

image

> In the middle column, the difference that will be most noticable is at the very high magnitude lateral acceleration. The new model is achiving 100% torque at lower magnitudes of lateral accel. Before, it was rounding off and having a flatter response at very high magnitude lateral accel.

image

> This is especially the case at 27-81mph, so it should hold curves better at mod/higher speeds now
> This also visible in the left columns, but is a bit easier to spot in the middle columns
> Interestingly, and somewhat contradictingly, the time-based lateral jerk response (i.e. based on past/present/future values of lateral accel) is now less gradual than before, which is opposite of the previous observation about the instantaneous lateral jerk. 

> I think the underlying for this slight "disagreement" in the gen 2 model vs gen 1 with respect to time-based vs instantaneous lateral jerk response is due to the data preprocessing script not perfectly time-aligning the different signals. If they were better aligned, then I would expect both types of lateral jerk response to evolve in the same way from gen 1 to gen 2.

wtogami:
This model has performed superbly for the past two weeks. I posted a PR because I wanted to make it easy for other people to test it with `dev-c3-new`. I didn't expect when it was marked as **draft** a week ago that would prevent merge in `dev-c3-new`.

I had intended to see if I could improve upon this with subsequent training. Four attempts produced a model worse than Gen 1 or unclear as to relative quality.

I want people to test this for a few more days while we think about it.

wtogami:
Subsequent attempts of making a better model have been inconclusive. It's time to merge this. It has been proven for weeks now.

wtogami:
This commit is unchanged since the beginning. Repushed by request of Kumar after the model was committed to the `neural-network-data` submodule.

rav4kumar (approved):

PR #958, opened 2025-05-30
ui: model fuzzy search

nayan8teen:


nayan8teen:
> I like this @nayan8teen but I dont know if there's anything we can do to also include the "shorthand" of the models. for example, North Dakota is also known as ndv2... It should be available on the model bunder under `shortName`.

@devtekve this is now supported.

devtekve (approved):

PR #971, opened 2025-06-01
Model panel: Backup and Restore liveDelay

Discountchubbs:


Discountchubbs:
@nayan8teen I haven't added any of these three toggles to the model panel.cc or header files yet. The ones are GasGating, LagdBackup, and LagdToggle , LagdToggle should default to "on" lagdBackup should probably be hidden if user disables lagdtoggle .

gas gating is self explanatory, but maybe we should figure out a user friendly title/description.


I can probably get to this later this week depending on work.

devtekve (approved):

PR #994, opened 2025-06-08
bug: fix missing mapd folder

royjr:


devtekve (dismissed):

devtekve:
We are pushing this for a few days on dev-c3 to check if it works as expected

PR #1017, opened 2025-06-26
fix: correctly establish NAT forwarding fw rule for tethering

bitnimble:
**Description**

Enabling the hotspot/tethering mode currently doesn't allow clients to actually tether off the Comma's SIM card connection; prior to 0.9.8 it did.

Upstream bug: https://github.com/commaai/openpilot/issues/34969
Discussion in SP discord: https://discord.com/channels/880416502577266699/1377315397564633169/1384735498379067454

This PR is a quick fix to restore functionality to pre-0.9.8 behaviour. Personally, I think that this should be fixed upstream, e.g. (adding the appropriate kernel modules, and adding it via nmcli instead), but that is a much more involved fix and I don't want to let perfect be the enemy of good - this works well for now, doesn't affect anything else, and is very easily removed once a better fix is in place.

**Verification**

Hotspot wasn't working for me; I cherry picked this commit to my own fork and have been running it for the past week and can confirm that hotspot is fixed for me. 

Credit to @cloudrainstar for the actual fix, I'm just the one raising it.

### DIFF
--- a/selfdrive/ui/qt/network/wifi_manager.cc
+++ b/selfdrive/ui/qt/network/wifi_manager.cc
@@ -490,6 +490,12 @@ void WifiManager::tetheringActivated(QDBusPendingCallWatcher *call) {
       qWarning() << "net.ipv4.ip_forward = 0";
       std::system("sudo sysctl net.ipv4.ip_forward=0");
     });
+  } else {
+    // nmcli can't set up NAT forwarding due to missing nf_tables kernel module, so we do it manually
+    QTimer::singleShot(5000, this, [=] {
+      qWarning() << "Setting up NAT forwarding for tethering";
+      std::system("sudo iptables-legacy -t nat -A POSTROUTING -s 192.168.43.0/24 -o wwan0 -j MASQUERADE");
+    });
   }
   call->deleteLater();
   tethering_on = true;


selfdrive/ui/qt/network/wifi_manager.cc
```diff
@@ -490,6 +490,12 @@ void WifiManager::tetheringActivated(QDBusPendingCallWatcher *call) {
       qWarning() << "net.ipv4.ip_forward = 0";
       std::system("sudo sysctl net.ipv4.ip_forward=0");
     });
+  } else {
```
bitnimble: oops, just realised that this wasn't in the sunnypilot overlays; should I wrap this in an #ifdef SUNNYPILOT?

devtekve: Well, it should really be on any of our overlays if we have one for wifi_manager. If we don't, then yeah, ifdef would be the most suitable IMHO.

devtekve:
The only thing I am not understanding is why it is working for me. So before merging this one I'd like to take one more look at how I made it working exactly on mine (i said it on the thread but it's been a bit since that)

selfdrive/ui/qt/network/wifi_manager.cc
bitnimble: Yeah, there's no existing overlay atm. will add an ifdef then 👌

bitnimble:
Yeah totally fair, lmk how it goes. would love to see if there's a way to permanently fix this for everyone without a code change needed 👍

bitnimble:
Bump on this one if possible!

I'm not sure why (repo perms?) but I can't change the source branch on this PR. I've merged some other things into that branch accidentally, and ideally this PR should point at my `bitnimble/fix-nat` branch instead... only commit on that branch is here: https://github.com/bitnimble/sunnypilot/commit/9bdd811706329484c99ddd55f2987447d8499604, so just ignore the diff on this PR. Or if someone with perms can update the source branch on this PR that would be stellar 🙏 

I've also done a completely fresh flash of AGNOS + SP and can confirm that this tethering issue still exists from a clean slate, which sounds the same as what Jakub was encountering [here](https://discordapp.com/channels/880416502577266699/1377315397564633169/1384179714247364661)

bitnimble:
Oh, I forgot about this PR - I recently rebased and tested, and unfortunately it's no longer working. I haven't debugged yet to see what the issue is, so please don't approve or merge this yet 😅

bitnimble:
Superceded by https://github.com/sunnypilot/sunnypilot/pull/1750; will close this one

PR #1018, opened 2025-06-26
feat: add toggle to persist tethering across reboots

bitnimble:
**Description**

Restores the feature of allowing the "Enable tethering" setting to persist after a reboot. This allows the hotspot to come up immediately when the Comma is turned on, without needing to manually go into `Network -> Advanced -> Enable tethering` every boot.

image

**Verification**

Compiled fork and tested on-device; confirmed that reboots instantly bring up the hotspot and client devices tether correctly.

### DIFF
--- a/selfdrive/ui/sunnypilot/qt/network/networking.cc
+++ b/selfdrive/ui/sunnypilot/qt/network/networking.cc
@@ -6,7 +6,7 @@
 NetworkingSP::NetworkingSP(QWidget *parent) : Networking(parent) {
   auto vlayout = wifiScreen->findChild<QVBoxLayout*>();
   auto hlayout = new QHBoxLayout();
-  
+
   // Create and setup scan button
   auto scanButton = new QPushButton(tr("Scan"));
   scanButton->setObjectName("scan_btn");
@@ -73,6 +73,7 @@ AdvancedNetworkingSP::AdvancedNetworkingSP(QWidget *parent, WifiManager *wifi) :
     QTimer::singleShot(2000, this, [=] {
       qDebug() << "Re-enabling tethering on startup";
       wifi->setTetheringEnabled(true);
+      tetheringToggle->setEnabled(true);
     });
   }
 }


selfdrive/ui/qt/network/networking.h
```diff
@@ -87,7 +87,11 @@ class AdvancedNetworking : public QWidget {
   void requestWifiScreen();
 
 public slots:
-  void toggleTethering(bool enabled);
+#ifdef SUNNYPILOT
+  virtual void toggleTethering(bool enabled);
```
bitnimble, line 91: Needed this to be overridable in the derived class, not sure if this is how we expect it to be done?

selfdrive/ui/sunnypilot/qt/network/networking.cc
```diff
@@ -43,3 +50,34 @@ NetworkingSP::NetworkingSP(QWidget *parent) : Networking(parent) {
...
+    params.putBool("PersistTetheringAcrossReboot", state);
+  });
+  // Insert it after tetheringToggle
+  auto buttons = findChildren<QPushButton*>();
+  auto tetheringToggleTitle = std::find_if(buttons.begin(), buttons.end(), [](QPushButton* btn) {
+    return btn->text() == tr("Enable Tethering");
+  });
+  auto tetheringToggle = tetheringToggleTitle != buttons.end() ? (*tetheringToggleTitle)->parentWidget() : nullptr; // Toggle widget is the parent of the button label
+  auto list = findChild<ListWidgetSP*>();
+  if (tetheringToggle && list) {
```
bitnimble, line 67: This was also kinda a bit weird in an attempt to avoid modifying the OP code as much as possible, might be cleaner with an #ifdef adding the qt object name / storing it as a protected var so we can access it here.

selfdrive/ui/sunnypilot/qt/network/networking.cc
```diff
@@ -43,3 +50,35 @@ NetworkingSP::NetworkingSP(QWidget *parent) : Networking(parent) {
...
+    list->AddWidgetAt(list->indexOf(tetheringToggle) + 1, tetheringPersistToggle);
+  }
+
+  // Initialize tethering on startup
+  if (params.getBool("TetheringEnabled") && params.getBool("PersistTetheringAcrossReboot")) {
+    QTimer::singleShot(2000, this, [=] {
+      qDebug() << "Re-enabling tethering on startup";
+      wifi->setTetheringEnabled(true);
+      tetheringToggle->setEnabled(true);
```
bitnimble, line 76: hm, this doesn't seem to work and I don't know Qt well enough to know why - afaik `tetheringToggle` is correctly targeting the `ToggleControlSP` (since it's correctly targeting it and adding our new toggle underneath it in the layout), is there some explicit re-render I need to call to get this to refresh visually?

sunnyhaibin:
No longer relevant as Qt is deprecated. Please open a new PR if it is still applicable.

PR #946, opened 2025-05-26
Learned lateral lag for torque extension and nnlc

twilsonco:
Three commits, could be split into two squash-merges. First two commits don't touch NNLC. Third commit is NNLC. Test changed in second commit should cover NNLC too.

PR #1027, opened 2025-07-06
[debug only] ui: `lateralDelay` + offset

wtogami:
UI to add optional offset on top of the learned `lateralDelay` while live learner is toggled on.

* Default is `0.0` seconds.
* Allow up to `0.20` offset maximum.
* The menu description is updated to display the component offset and total delay.

## Summary by Sourcery

Introduce an optional lateral delay offset on top of the live-learned steer delay and expose it in the settings UI, updating delay calculations and menu descriptions accordingly.

New Features:
- Add a UI control to configure the learned delay offset when live learning is enabled.

Enhancements:
- Incorporate the offset into the live delay computation and update the description string to show the learner delay, offset, smoothing filter, and total delay.

Chores:
- Register the new `LagdToggledelayoffset` parameter key and set its default value in the manager initialization.

### DIFF
--- a/common/params_keys.h
+++ b/common/params_keys.h
@@ -136,6 +136,7 @@ inline static std::unordered_map<std::string, uint32_t> keys = {
     {"CustomAccShortPressIncrement", PERSISTENT | BACKUP},
     {"DeviceBootMode", PERSISTENT | BACKUP},
     {"EnableGithubRunner", PERSISTENT | BACKUP},
+    {"InteractivityTimeout", PERSISTENT | BACKUP},
     {"MaxTimeOffroad", PERSISTENT | BACKUP},
     {"Brightness", PERSISTENT | BACKUP},
     {"ModelRunnerTypeCache", CLEAR_ON_ONROAD_TRANSITION},
--- a/selfdrive/ui/sunnypilot/qt/offroad/settings/device_panel.cc
+++ b/selfdrive/ui/sunnypilot/qt/offroad/settings/device_panel.cc
@@ -87,6 +87,17 @@ DevicePanelSP::DevicePanelSP(SettingsWindowSP *parent) : DevicePanel(parent) {
     params.put("DeviceBootMode", QString::number(index).toStdString());
     updateState();
   });
+
+  interactivityTimeout =  new OptionControlSP("InteractivityTimeout", tr("Interactivity Timeout"),
+                                     tr("Apply a custom timeout for settings UI."
+                                        "\nThis is the time after which settings UI closes automatically if user is not interacting with the screen."),
+                                     "", {0, 120}, 10, true, nullptr, false);
+
+  connect(interactivityTimeout, &OptionControlSP::updateLabels, [=]() {
+    updateState();
+  });
+
+  addItem(interactivityTimeout);
   
   // Brightness
   brightness = new Brightness();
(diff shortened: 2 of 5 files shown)


devtekve:
Im not sure mixing live learning + offset would be good. Live learning would likely be compensating for the offset. If anything, a PR should be made upstream to improve the live learning

wtogami:
> Live learning would likely be compensating for the offset.

Ooh you mean the `lateralDelay` estimate would change **because the offset changes subsequent live measurements**?

wtogami:
Marking this PR '[debug only]` to keep it around as it revealed some interesting behavior.

Ever since #946 merged I have been experiencing some type of regression on my 2023 Toyota RAV4 Prime. Previously TR7+NNLC was extremely reliable. It was always centered in lane and it had a "locked in" feeling. After #946 it became unreliable. It would no longer stay stable in lane. Intermittently it would get REAL BAD with lane excursions.

#946 was a big change. Previously NNLC was adding a hard-coded 0.3 seconds to Toyota `steeringActuatorDelay` 0.12. 0.42 seconds hard-coded became live delay estimates in the range of 0.22-0.25 seconds. I previously thought the regressed behavior was because the live delay estimate was too low (not true).

https://github.com/wtogami/openpilot/commit/daef038debe3b834a5d2166636c5be97f92eb49e
Earlier version of #946 had NNLC use `lateralDelay` plus an offset. I previously tested NNLC with `lateralDelay` + `0.1`. It felt better than `lateralDelay` alone as merged into `master-new` but I now realize it is much better to disable lagd. 

I tested NNLC with + `0.1` for weeks. That did not seem to affect the `lateralDelay` estimate. Granted the NNLC `LATERAL_LAG_MOD` is not the same thing as this offset PR. That modifier affected only NNLC plan-ahead intervals and steering while this PR affects everything (NNLC, torqued, modeld, etc.) I have no idea how that offset differs from this offset.

This offset PR revealed something interesting. I'm seeing exactly the same misbehavior at offsets `0.0`, `0.02` and `0.05`. Meanwhile disabling live delay and choosing a low fixed delay roughly equal to `lateralDelay` eliminated the misbehavior entirely.

It seems something behaves poorly after there is a big change to the `lateralDelay` estimate during a drive. My first guess was NNLC's plan-ahead intervals would be confused by the interval length suddenly changing. twilsonco said however the most it would be effected would be for 0.3 seconds. His guess of what else might be affected by a sudden `lateralDelay` change: "Lateral delay is also passed into the driving model. Who knows how that plays out. It's sort of stateful since it gets the previous and current frame data, so there could be some weird resonance back-and forth behavior that starts when lagd changes suddenly."

So this PR might not be suitable for merge (as it doesn't seem to fix anything) but people might find it interesting to help debug other problems.

Discountchubbs:
@wtogami:

I do enjoy your methodology used in attempt to validate your research hypothesis here. Because of this, we can expand this testing criteria to more users by having a prebuilt made for you. The prebuilt will be name of branch with `-prebuilt` as the suffix.

wtogami:
Here is my attempt to verify lagd estimates are not adversely influenced by the offset.

15 miles drive mostly highway with `lateralDelay` + `0.20` offset. Here is how the estimate changed between segments.
* 0.228 (four segments)
* 0.225
* 0.222 (several of this)
* 0.223
* 0.222
* 0.226
* 0.227

Estimates changed a tiny bit during that drive. This confirms the offset does NOT throw off the estimator.

wtogami:
When my `lateralDelay` estimate is in this low range near `0.22` my Toyota suffers from a noticeable amount of oversteer while using exactly `lateralDelay`. Adding offset `0.03` to that is usually enough to correct it.

Similarly Live Disable disabled -> `steerActuatorDelay` + `0.10` adds up to `0.22`. The same oversteer is present then. NNLC works under this condition. It just feels bad. Setting fixed software delay to `0.13` seems enough to correct this too.

Unfortunately the `lateralDelay` estimate is often `0.25` or sometimes even `0.27`. Also there seems to be a mystery undesired behavior (likely from modeld) where its ability to stay in lane becomes unstable when the estimate has big changes during a drive. For these reasons it seems disabling live delay and setting a high-enough fixed delay seems to be the most stable way to use it for now.

wtogami:
I will likely bring this experiment back in sync at some point later. Closing for now.

PR #1033, opened 2025-07-10
Visuals: Lead Chevron Info 

rav4kumar:
- [x] add individual params

## Summary by Sourcery

Add configurable metric overlays below lead vehicle chevrons to show distance, speed, and time-to-contact with fade animations.

New Features:
- Display dynamic distance, speed, and time-to-contact metrics under lead chevrons with fade-in/out effects.
- Expose new ChevronInfo parameter in the Visuals settings panel to configure which metrics are shown.
- Initialize ChevronInfo in manager defaults and mark it as persistent for parameter storage.

Enhancements:
- Extend ModelRenderer with drawLeadStatus and drawLeadStatusAtPosition to render metric overlays.
- Update VisualsPanel UI to include the ChevronInfo toggle control and refresh logic.

### DIFF
--- a/system/manager/manager.py
+++ b/system/manager/manager.py
@@ -56,6 +56,7 @@ def manager_init() -> None:
     ("CustomAccLongPressIncrement", "5"),
     ("CustomAccShortPressIncrement", "1"),
     ("DeviceBootMode", "0"),
+    ("DisableUpdates", "0"),
     ("DynamicExperimentalControl", "0"),
     ("HyundaiLongitudinalTuning", "0"),
     ("InteractivityTimeout", "0"),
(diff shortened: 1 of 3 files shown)


system/manager/manager.py
```diff
@@ -50,6 +50,8 @@ def manager_init() -> None:
     ("BlindSpot", "0"),
     ("BlinkerMinLateralControlSpeed", "20"),  # MPH or km/h
     ("BlinkerPauseLateralControl", "0"),
+    ("Brightness", "0"),
+    ("ChevronInfo", "0"),
```
devtekve:
```suggestion
    ("ChevronInfo", "4"),
```
Defaulting it to All and let's let the users change it, otherwise they may not even know there's a new feature. And it's hella cool!

devtekve (dismissed):

devtekve:
10/10 for the feature! 

however I don't love the fact that it's not on a sunnypilot module, and instead is on model.c from upstream which may give us issues syncing. However, because of the migration to raylib, it would just make us waste time so we are moving along with this

Candy0707:
The text will be cut when it exceeds (100 km/h)

PR #1061, opened 2025-07-19
backup: Adding more keys to be backed up

devtekve:


Discountchubbs (dismissed):
This adds more keys to be backed and restored up during backup settings on sunnylink panel

PR #1064, opened 2025-07-21
pyui: add sp toggles

nayan8teen:
## Summary by Sourcery

Enable optional sunnypilot-styled toggles by introducing a new persistent sunnypilot_ui parameter, updating settings layouts to conditionally use SP toggle items, and adding sunnypilot-specific list view, toggle, and style libraries.

New Features:
- Add persistent sunnypilot_ui parameter with default enabled in Params keys and manager defaults
- Implement sunnypilot-specific ToggleSP and ListItemSP components with custom rendering and styles
- Conditionally import and use toggle_item_sp in developer and settings layouts when sunnypilot_ui is enabled


selfdrive/ui/layouts/settings/developer.py
```diff
@@ -4,6 +4,9 @@
 from openpilot.common.params import Params
 from openpilot.selfdrive.ui.widgets.ssh_key import ssh_key_item
 
+if Params().get_bool("sunnypilot_ui"):
+  from openpilot.system.ui.sunnypilot.lib.list_view import toggle_item_sp as toggle_item
+
```
devtekve, line 9: Niiiiiice!

common/params_keys.h
```diff
@@ -123,6 +123,8 @@ inline static std::unordered_map<std::string, uint32_t> keys = {
     {"Version", PERSISTENT},
 
     // --- sunnypilot params --- //
+    {"sunnypilot_ui", PERSISTENT},
```
devtekve:
```suggestion
    {"sunnypilot_ui", PERSISTENT | BACKUP},
```

system/manager/manager.py
```diff
@@ -45,6 +45,8 @@ def manager_init() -> None:
   ]
 
   sunnypilot_default_params: list[tuple[str, str | bytes]] = [
+    ("sunnypilot_ui", "1"), # Use sunnypilot UI for raylib. Disable to use stock.
```
devtekve: This is making it the default, careful if this param is used to choose between QT and raylib

devtekve (dismissed):
looking great. I didn't go into the specific details of the components themselves, the organization looks good and seems flexible enough for us to change it down the road should we need to. Good starting point

system/manager/manager.py
nayan8teen: Nah, this is basically an equivalent of the SUNNYPILOT compile parameter we have on the cpp side. That's why defaulting it to 1. If needed, later on we could add something along the lines of "if compile param SUNNYPILOT, then set this to 1 else 0" 

I have a different param I'm using for the switch between qt & raylib - but that's on my branch only right now.

nayan8teen:
will reopen with updated when ready to merge

PR #1066, opened 2025-07-22
Visual: Move chevron to sp/qt

rav4kumar:


devtekve:
Just a reminder: let’s make sure the info is only displayed when long is active, regardless of the setting of the info. Otherwise the lead info may still be shown in the screen out of place

selfdrive/ui/sunnypilot/qt/onroad/model.cc
```diff
@@ -48,5 +48,143 @@ void ModelRendererSP::drawPath(QPainter &painter, const cereal::ModelDataV2::Rea
...
+                text_color = QColor(80, 255, 120, (int)(255 * lead_status_alpha));
+            }
+        }
+
+        painter.setPen(text_color);
+        painter.drawText(rect, Qt::AlignCenter, text_lines[i]);
+    }
+
+    painter.setPen(Qt::NoPen);
+}
```
devtekve:
```suggestion
}

```

devtekve (changes requested):
This is waiting to have the lead info disabled with there's no long control and / or if the chevron is not visible. (the goal is to not display this when chevron is not visible)

rav4kumar:
> This is waiting to have the lead info disabled with there's no long control and / or if the chevron is not visible. (the goal is to not display this when chevron is not visible)

done!

selfdrive/ui/sunnypilot/qt/onroad/model.cc
```diff
@@ -86,5 +86,154 @@ void ModelRendererSP::drawPath(QPainter &painter, const cereal::ModelDataV2::Rea
...
+    bool longitudinal_control = sm["carParams"].getCarParams().getOpenpilotLongitudinalControl();
+    if (!longitudinal_control) {
+        lead_status_alpha = std::max(0.0f, lead_status_alpha - 0.05f);
+        return;
+    }
+
+    if (!sm.alive("radarState")) {
+        lead_status_alpha = std::max(0.0f, lead_status_alpha - 0.05f);
+        return;
+    }
```
sunnyhaibin: We should also block users from enabling the toggle if they do not have longitudinal available in settings. We do this for certain longitudinal options/toggles already: https://github.com/sunnypilot/sunnypilot/blob/810a2d9448df89ef84ecc1c794b63fad55350d69/selfdrive/ui/sunnypilot/qt/offroad/settings/longitudinal_panel.cc#L37-L48

selfdrive/ui/sunnypilot/qt/onroad/model.h
```diff
@@ -17,6 +17,17 @@ class ModelRendererSP : public ModelRenderer {
   void update_model(const cereal::ModelDataV2::Reader &model, const cereal::RadarState::LeadData::Reader &lead) override;
   void drawPath(QPainter &painter, const cereal::ModelDataV2::Reader &model, const QRect &rect) override;
 
+  // Lead status display methods
+  void drawLeadStatus(QPainter &painter, int height, int width);
+  void drawLeadStatusAtPosition(QPainter &painter,
+                               const cereal::RadarState::LeadData::Reader &lead_data,
+                               const QPointF &chevron_pos,
+                               int height, int width,
+                               const QString &label);
```
sunnyhaibin:
```suggestion
  void drawLeadStatusAtPosition(QPainter &painter,
                                const cereal::RadarState::LeadData::Reader &lead_data,
                                const QPointF &chevron_pos,
                                int height, int width,
                                const QString &label);
```

selfdrive/ui/sunnypilot/qt/onroad/model.cc
sunnyhaibin, line 1: 2-space indents to match the rest of the style.

selfdrive/ui/sunnypilot/qt/onroad/model.cc
```diff
@@ -86,5 +86,139 @@ void ModelRendererSP::drawPath(QPainter &painter, const cereal::ModelDataV2::Rea
   } else {
     // Normal path rendering
     ModelRenderer::drawPath(painter, model, surface_rect.height());
+
+    drawLeadStatus(painter, surface_rect.height(), surface_rect.width());
```
sunnyhaibin: This is only being drawn when the rainbow path is off. What's reason for this?

selfdrive/ui/sunnypilot/qt/onroad/model.cc
```diff
@@ -86,5 +86,139 @@ void ModelRendererSP::drawPath(QPainter &painter, const cereal::ModelDataV2::Rea
...
+                                               const QPointF &chevron_pos,
+                                               int height, int width,
+                                               const QString &label) {
+  float d_rel = lead_data.getDRel();
+  float v_rel = lead_data.getVRel();
+  auto *s = uiState();
+  auto &sm = *(s->sm);
+  float v_ego = sm["carState"].getCarState().getVEgo();
+
+  int chevron_data = std::atoi(Params().get("ChevronInfo").c_str());
```
sunnyhaibin: We should have these checked under `ui.cc`, instead of creating a new instance of Params on every call. https://github.com/sunnypilot/sunnypilot/blob/3df1b53fab4341074ace53b2797eb16eb5ed3e09/selfdrive/ui/sunnypilot/ui.cc#L61-L75

PR #1093, opened 2025-07-30
ci: use self-hosted x64

sunnyhaibin:


devtekve:
@sunnyhaibin not sure if we really need to worry about this or if we should close this PR and wait until we have a need to reassess?

sunnyhaibin:
Don't think we need this anytime soon. Closing.

PR #1102, opened 2025-08-01
mapd: suppress output

royjr:
## Summary by Sourcery

Enhancements:
- Silence mapd stdout and stderr by launching it through a bash command with output redirection


devtekve (approved):
LGTM

PR #1113, opened 2025-08-03
NNLC: fix friction not being applied to error

sunnyhaibin:

### DIFF
--- a/sunnypilot/selfdrive/controls/lib/nnlc/nnlc.py
+++ b/sunnypilot/selfdrive/controls/lib/nnlc/nnlc.py
@@ -8,6 +8,7 @@ from collections import deque
 import math
 import numpy as np
 
+from opendbc.car import FRICTION_THRESHOLD, get_friction
 from opendbc.car.interfaces import LatControlInputs
 from openpilot.common.filter_simple import FirstOrderFilter
 from openpilot.common.params import Params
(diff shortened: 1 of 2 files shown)


sunnyhaibin:
Hi @twilsonco - can you please see if this change makes sense? This is changed due to an upstream change of obtaining friction:
- https://github.com/commaai/opendbc/pull/2512

sunnypilot/selfdrive/controls/lib/nnlc/nnlc.py
```diff
@@ -126,7 +127,7 @@ def update_neural_network_feedforward(self, CS, params, calibrated_pose) -> None
     self._ff = self.model.evaluate(nn_input)
 
     # apply friction override for cars with low NN friction response
-    # TODO-SP: verify with twilsonco if this change is appropriate
     if self.model.friction_override:
       self._pid_log.error += self.torque_from_lateral_accel(LatControlInputs(0.0, 0.0, CS.vEgo, CS.aEgo), self.torque_params,
```
twilsonco: This line should be removed. (It does nothing anyhow since the input lat accel is zero and there's no roll compensation or lateral jerk input either.)

twilsonco:
Other than the line I commented on, looks good.

sunnyhaibin:
@twilsonco thanks for the review!

PR #1117, opened 2025-08-03
flash: skip if deprecated panda is detected

sunnyhaibin:


devtekve (approved):
It’s a shame we have To do this but i guess hands are tied

PR #1122, opened 2025-08-05
hyundai: Modify Tune and Allow FCW

Discountchubbs:
When a FCW alert is triggered, this new path completely overrides any, and all custom/stock tuning to allow higher likelihood of successful braking.

This in accordance with ISO 15622:2018 allows, `Braking jerk`: Significant short-period changes of the actual acceleration by introducing a short pulse to the brakes system.

As this is not AEB, this system does not allow input of first 3 m/s2 then 5 m/s2 deceleration, and only allows jerk to be "charged" allowing for quicker deceleration at -3.5 m/s^2.

> [!NOTE]
> opendbc PR:
> * https://github.com/sunnypilot/opendbc/pull/212


sunnyhaibin:
@Discountchubbs Do we need this PR since all changes should be on the opendbc repo?

PR #1127, opened 2025-08-06
params: fix auto type cast on put

sunnyhaibin:


devtekve:
Let’s validate backup manager with this since this is the first time we have json params backed up afaik

PR #1132, opened 2025-08-08
ci: adjusting build-single to comply with githubs change

Discountchubbs:
github changed the way number inputs are used. now number inputs are floats instead of just integers. So what was an input of 8, is now 8.0

## Summary by Sourcery

CI:
- Change the input types for recompiled_dir, json_version, generation, and version from number to string in the build-single-tinygrad-model workflow

### DIFF
--- a/.github/workflows/build-single-tinygrad-model.yaml
+++ b/.github/workflows/build-single-tinygrad-model.yaml
@@ -42,11 +42,11 @@ on:
       recompiled_dir:
         description: 'Existing recompiled directory number (e.g. 3 for recompiled3)'
         required: true
-        type: number
+        type: string
       json_version:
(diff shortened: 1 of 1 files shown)


Discountchubbs:
https://github.com/sunnypilot/sunnypilot/actions/runs/16838924877/job/47705140330
Screenshot 2025-08-08 at 3 23 36 PM

https://github.com/orgs/community/discussions/67182

This was what I could find..

Apparently it is a bug that they never fixed and can happen sometimes. Regardless, string for these will be more formidable for int format anyways to prevent this from happening again. There was an update to github workflows engine today 8 AUG 2025, could that have brought back the bug, possibly but thats not something we need to worry about.

PR #1139, opened 2025-08-10
models panel ui: live description for lagd toggle

Discountchubbs:
> [!NOTE]
> DEPENDS ON:
> * https://github.com/sunnypilot/sunnypilot/pull/1138


Discountchubbs:
https://github.com/user-attachments/assets/11d4a4a9-aaf7-48ed-815d-cc483a67e64f

wtogami:
While live delay is enabled please display the estimate to the thousandths place. The old code previously displayed to the thousandths place. That was helpful to be able to see the fact that the estimate is changing. For example 0.211 and 0.219 is the kind of difference I wanted to see while debugging. Currently that would be hidden.

wtogami:
I was requested to submit #1163 against this PR. The new github rules auto-closed it because it wasn't submitted against `master`.

PR #1142, opened 2025-08-11
NNLC: cleanup unused attributes

sunnyhaibin:

### DIFF
--- a/sunnypilot/selfdrive/controls/lib/latcontrol_torque_ext_base.py
+++ b/sunnypilot/selfdrive/controls/lib/latcontrol_torque_ext_base.py
@@ -11,7 +11,8 @@ from openpilot.selfdrive.controls.lib.drive_helpers import CONTROL_N
 from openpilot.selfdrive.modeld.constants import ModelConstants
 
 LAT_PLAN_MIN_IDX = 5
-LATERAL_LAG_MOD = 0.0 # seconds, modifies how far in the future we look ahead for the lateral plan
+LATERAL_LAG_MOD = 0.0  # seconds, modifies how far in the future we look ahead for the lateral plan
+
 
 def get_predicted_lateral_jerk(lat_accels, t_diffs):
   # compute finite difference between subsequent model_v2.acceleration.y values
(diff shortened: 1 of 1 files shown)


sunnyhaibin:
@twilsonco can you please review?

sunnypilot/selfdrive/controls/lib/latcontrol_torque_ext_base.py
```diff
@@ -118,7 +117,7 @@ def update_calculations(self, CS, VM, desired_lateral_accel):
       desired_lateral_jerk = (np.interp(self.desired_lat_jerk_time, ModelConstants.T_IDXS,
                               self.model_v2.acceleration.y) - desired_lateral_accel) / self.desired_lat_jerk_time
       self.lookahead_lateral_jerk = get_lookahead_value(predicted_lateral_jerk[LAT_PLAN_MIN_IDX:friction_upper_idx], desired_lateral_jerk)
-      if not self.use_steering_angle or self.lookahead_lateral_jerk == 0.0:
+      if self.lookahead_lateral_jerk == 0.0:
```
twilsonco, line 120: Explanation of this logic: When on straights, there will be no desired lateral jerk that persists long enough to be considered "deliberate", i.e. lookahead lateral jerk will be zero. In that case, we raise the lat acceleration error-based component of friction compensation to full (`self.lat_accel_friction_factor = 1.0`) so that we get better centering behavior. (On straights there is no non-zero desired steering angle, so there is no feedforward component to steer torque. That is, there can only be error-driven steer torque.)

sunnypilot/selfdrive/controls/lib/latcontrol_torque_ext_base.py
```diff
@@ -118,7 +117,7 @@ def update_calculations(self, CS, VM, desired_lateral_accel):
...
         self.lookahead_lateral_jerk = 0.0
```
twilsonco: Since we're cleaning up. This line can be removed, as we already know that `self.lookahead_lateral_jerk == 0.0`.

twilsonco (commented):
This looks good

sunnyhaibin:
> This looks good

Thanks!

PR #1146, opened 2025-08-11
car: initialize brand-related settings in opendbc

sunnyhaibin:
## Summary by Sourcery

Introduce a generic brand-specific parameter initialization flow and wire it into car setup, replacing the old custom longitudinal tuning logic.

Enhancements:
- Add initialize_params function to extract brand-related settings from Params
- Pass the initialized parameter list into get_car during Car instantiation
- Remove the legacy _initialize_custom_longitudinal_tuning function and its setup call


Discountchubbs (dismissed):
merge latest master changes but its approved

PR #1150, opened 2025-08-12
sunnypilot modeld: refactor gen12 parsing 

Discountchubbs:


Discountchubbs:
Hmm this may conflict with our existing simple plan model. will need to check that

sunnyhaibin (approved):

PR #1129, opened 2025-08-06
bugfix: fix default params

nayan8teen:

PR #1154, opened 2025-08-13
Fix "Default" interactivity timeout display

wtogami:
If param `InteractivityTimeout` does not exist, timeoutValue == "", resulting in the Interactivity Timeout UI displaying "s" by default.

Fix displayed "Default" by checking for "0" or "".

# Before
image

# After
image

## Summary by Sourcery

Bug Fixes:
- Show "Default" label for interactivity timeout when the parameter is empty or "0" instead of displaying "s"

### DIFF
--- a/selfdrive/ui/sunnypilot/qt/offroad/settings/device_panel.cc
+++ b/selfdrive/ui/sunnypilot/qt/offroad/settings/device_panel.cc
@@ -211,7 +211,7 @@ void DevicePanelSP::updateState() {
   toggleDeviceBootMode->setDescription(deviceSleepModeDescription(currStatus));
 
   QString timeoutValue = QString::fromStdString(params.get("InteractivityTimeout"));
-  if (timeoutValue == "0") {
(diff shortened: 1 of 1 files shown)


nayan8teen:
@wtogami even though this would fix this particular issue, the root cause is how upstream changed the behavior of params, specially in terms of default values.

We have a PR to fix the default param behavior in general - which would fix this issue too. It needs to be updated with a slightly different approach though. Should be updated and fixed soon.

https://github.com/sunnypilot/sunnypilot/pull/1129

wtogami:
Thanks. Closing.

nayan8teen:
> Thanks. Closing.

@wtogami This has been reopened. 

We discussed and decided to do both - the other PR to fix backwards compatibility, but also be more explicit about param defaults/None values going forward, like you have done here. Your PR would basically be a step towards this explicit cleanup. 

Also, when I say discussed, I mean @devtekve took me into a corner and made me do this 😏🤣

PR #1186, opened 2025-08-28
agnos: split launch for c3 and c3x to support custom agnos

devtekve:
Since C3 has been deprecated, we will need to find a way to keep it up to date. However, until we can make our own agnos reliably, we need to prevent them from accidentally being updated and potentially getting bricked.

## Summary by Sourcery

Prevent AGNOS updater from running on deprecated C3 devices by detecting the device model and skipping updates for models ending in "tici"

Bug Fixes:
- Prevent AGNOS updates on deprecated C3 devices by skipping updates for models matching "tici"

Enhancements:
- Read the device model from the firmware device tree in launch_chffrplus.sh
- Wrap the AGNOS update logic in a case statement to apply it conditionally


wtogami:
This would prevent C3 from upgrading to the final version of C3 AGNOS. Perhaps hard-code upgrading to the final C3 AGNOS here instead?

devtekve:
> This would prevent C3 from upgrading to the final version of C3 AGNOS. Perhaps hard-code upgrading to the final C3 AGNOS here instead?

redid the approach, complete split for c3 and c3x for the launch so that they can be independently maintained if we need to.

PR #1191, opened 2025-08-30
Reorder README tables: show -new branches first

michael-was-taken:
This PR updates the README installation guide:

Moved the -new branches table above the older branches table so that the most relevant branches are shown first.
Added a tip noting that users can install with sunnypilot/staging-c3-new.


michael-was-taken:
Added the legacy branches header as suggested by DevTek.

devtekve (approved):
Thank you very much!

PR #1192, opened 2025-08-30
#define HudRenderer HudRendererSP

Candy0707:
Fixed sunnypilot/ui/qt/hud not being referenced correctly

## Summary by Sourcery

Fix HUD renderer references in Sunnypilot UI by mapping HudRenderer to HudRendererSP, adjusting header includes, and updating the HudRendererSP constructor to explicitly call and inherit from HudRenderer.

Enhancements:
- Define HudRenderer as HudRendererSP under the SUNNYPILOT flag to ensure correct class resolution
- Move the include of model.h into the appropriate conditional branch in annotated_camera.h
- Add an explicit base class initializer call in HudRendererSP’s constructor and mark it explicit


selfdrive/ui/qt/onroad/annotated_camera.h
```diff
@@ -12,9 +11,11 @@
 #include "selfdrive/ui/sunnypilot/qt/onroad/model.h"
 #define ExperimentalButton ExperimentalButtonSP
 #define ModelRenderer ModelRendererSP
+#define HudRenderer HudRendererSP
 #else
 #include "selfdrive/ui/qt/onroad/buttons.h"
 #include "selfdrive/ui/qt/onroad/hud.h"
+#include "selfdrive/ui/qt/onroad/model.h"
```
devtekve, line 18: why moving this? everything works fine with it above and the HudRendererSP is called...

devtekve (changes requested):
need clarification to make sure I ain't seeing things wrong

selfdrive/ui/qt/onroad/annotated_camera.h
Candy0707: The SunnyPilot model has been overwritten.
So the reference is
#include "selfdrive/ui/sunnypilot/qt/onroad/model.h"

I'm adding some new elements to the HUD, so I'm using HudRendererSP

Candy0707:
> need clarification to make sure I ain't seeing things wrong
Yes, there is no mistake，Of course you don't have to move.

Candy0707:
image
Here is what I'm testing

PR #1203, opened 2025-09-02
Allows us to stat getting the swaglogs from devices (same as comma's)…

devtekve:


sunnyhaibin:
Marking as Draft for now.

sunnyhaibin:
#1580

PR #1204, opened 2025-09-03
bugfix: streamline LiveDelay parameter loading with safe handling

devtekve:


Discountchubbs (approved):
exception handling to allow safe sunnylink restore

PR #1210, opened 2025-09-04
update: remove git cleanup in finalized stage

sunnyhaibin:
## Summary by Sourcery

Enhancements:
- Eliminate git gc and git lfs prune operations during finalized updates


sunnyhaibin (commented):
We will monitor for any significant regressions with this change and revert if necessary.

PR #1244, opened 2025-09-14
bugfix: param store to support the latest param changes

devtekve:
> [!NOTE]
> * Sister PR: https://github.com/sunnypilot/opendbc/pull/273

This PR fixes the param store to properly fetch and return data with the new format, and allows the users (in opendbc) to get the param with the actual expected data type.

## Summary by Sourcery

Fix the parameter store to fetch and return parameters in the new schema and with proper types by updating the Cap'n Proto definitions, refactoring the store logic, and enhancing the utility functions

New Features:
- Extend CarControlSP.Param schema to include a binary Data value and ParamType enum for typed parameters

Bug Fixes:
- Fix ParamStore.update to correctly retrieve and cache parameters in the latest format with expected data types

Enhancements:
- Refactor ParamStore to use an internal dict of custom.CarControlSP.Param and expose them via a param_list property
- Enhance get_param_as_byte to accept an existing Params instance and delegate type conversion to a new helper
- Update controlsd_ext to consume the new param_list instead of the old publish method

### DIFF
--- a/sunnypilot/sunnylink/utils.py
+++ b/sunnypilot/sunnylink/utils.py
@@ -60,9 +60,9 @@ def get_api_token():
   print(f"API Token: {token}")
 
 
-def get_param_as_byte(param_name: str) -> bytes | None:
+def get_param_as_byte(param_name: str, params=None) -> bytes | None:
   """Get a parameter as bytes. Returns None if the parameter does not exist."""
-  params = Params()
+  params = params or Params() # Use existing Params instance if provided
   param = params.get(param_name)
   if param is None:
     return None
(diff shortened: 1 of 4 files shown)


sunnypilot/sunnylink/utils.py
```diff
@@ -85,6 +85,16 @@ def save_param_from_base64_encoded_string(param_name: str, base64_encoded_data:
...
 
+  # We convert to string anything that isn't bytes first. We later transform further.
+  param_value = _convert_param_to_type(value, param_type)
+  params.put(param_name, param_value)
+  
+def _convert_param_to_type(value: bytes, param_type: ParamKeyType) -> bytes | str | int | float | bool | dict | None:
+  """ 
+  Convert a byte value to the specified param type. Used internally when getting a Param to convert it to the right type.
+  If this method looks familiar, it's because on SP we have a similar one in openpilot/sunnypilot/car/__init__.py.
+  """
```
devtekve: The diff here looks unreadable on GH. So I will explain: 

split the "conversion" to it's own method and used it. The reason for the split is so that it's obvious for us theres "duplicated" code between this and opendbc. I don't have a better place to put this that could make it usable by both opendbc and sp, given that "params" is a concept that really only exists on SP, moving it to opendbc would feel wrong. And importing SP from opendbc is not a valid approach either

PR #1247, opened 2025-09-14
Honda Clarity: brake hold fix

mvl-boston:
Honda Clarity brake hold fix - generalized for Nidec hybrids

Dependent on: https://github.com/sunnypilot/opendbc/pull/314/files

Todo:
- [x] Drive test

## Summary by Sourcery

Prevent unintended system resume on Nidec-based hybrid vehicles (e.g., Honda Clarity) when brake hold is active by generating a belowEngageSpeed event.

Bug Fixes:
- Add check for CS.brakeHoldActive with CS.blockPcmEnable to trigger belowEngageSpeed and block engagement during brake hold

Enhancements:
- Generalize brake hold handling logic for all Nidec hybrid platforms instead of Honda Clarity only


devtekve:
@mvl-boston once you test drive don't forget to upload the route logs and make them public or share with us with our email so we can validate

mvl-boston:
Validation route log below:
3fef3f1f530d0ff6/0000000b--4c541a812b

sunnyhaibin (commented):
- See https://github.com/sunnypilot/opendbc/pull/276#pullrequestreview-3294841064
- `EventName.belowEngageSpeed` is a `WARNING` alert and it does not block engagement. This event seems misleading as it notifies users that they cannot engage ACC while traveling below the documented set speed, but this is mainly a disallowed engagement due to Brake Hold. If this is mainly for notification purposes and not for actively blocking controls, another variant of `EventName.brakeHold` would probably be a better option.

mvl-boston:
> * See [sunnypilot/opendbc#276 (review)](https://github.com/sunnypilot/opendbc/pull/276#pullrequestreview-3294841064)
> 
>     * `EventName.belowEngageSpeed` is a `WARNING` alert and it does not block engagement. This event seems misleading as it notifies users that they cannot engage ACC while traveling below the documented set speed, but this is mainly a disallowed engagement due to Brake Hold. If this is mainly for notification purposes and not for actively blocking controls, another variant of `EventName.brakeHold` would probably be a better option.

belowEngageSpeed is an ET.NO_ENTRY alert in events.py.   This causes the state machine to cancel the attempted cruise enable.  The driver can then press the gas and resume cruise instead which is what the car permits.

sunnyhaibin (approved):
Thanks for the PR!

PR #1252, opened 2025-09-16
UI: Driving Screen Off

### LINKED ISSUE #774: Device: Driving Screen Off: Wake with non-critical events

nayan8teen:
Closes #774


devtekve:
Lets make sure we move: 

* Interactivity timeout
* Brightness

from `Device` panel to `Display`


(awesome work!)

devtekve:
* brightness `-` button doesn't go disabled when on "dark" but pressing it does nothing so it must be the end of the available setings

devtekve:
* "timeout" and "brightness" make sense once you gave it some thought but the first impression is that it isn't clear what each does, and specially timeout being "always on" 
* The toggle is a bit confusing itself, because we'd like to always have brigtness control ? 

One thought I am having as I write this is "maybe we just need a toggle that says keep screen off unless alert" and intuitively, the regular brightesss set would already be used (whatever you manually chosen, they may not need to be side by site if that makes sense)

nayan8teen:
> * brightness `-` button doesn't go disabled when on "dark" but pressing it does nothing so it must be the end of the available setings

Fixed.

nayan8teen:
> * "timeout" and "brightness" make sense once you gave it some thought but the first impression is that it isn't clear what each does, and specially timeout being "always on"
> * The toggle is a bit confusing itself, because we'd like to always have brigtness control ?
> 
> One thought I am having as I write this is "maybe we just need a toggle that says keep screen off unless alert" and intuitively, the regular brightesss set would already be used (whatever you manually chosen, they may not need to be side by site if that makes sense)

Makes more sense this way?

image

sunnyhaibin (approved):
LGTM. Awesome!!!

PR #1256, opened 2025-09-18
Speed Limit: Resolver

### LINKED ISSUE #1255: Speed Limit: Resolver
#802

sunnyhaibin:
Closes #1255

## Summary by Sourcery

Add speed limit resolution feature: implement a SpeedLimitResolver that merges car and map speed limit sources per user-configurable policy and offset, integrate it into the SpeedPlanner pipeline and messaging schema, and expose settings in the offroad UI

New Features:
- Implement SpeedLimitResolver to compute speed limits from car and map data according to configurable policies and offsets
- Integrate resolver output into LongitudinalPlannerSP and publish speedLimit data in the plan message
- Add a Speed Limit settings panel in the offroad longitudinal settings UI for source policy and offset configuration
- Extend Cap’n Proto schemas to include speedLimit and resolver fields in LongitudinalPlanSP and CarStateSP

Enhancements:
- Enhance back button styling in Longitudinal settings panel

Build:
- Update params_keys with new speed limit parameters
- Include new UI source files in SConscript and adjust plannerd subscriptions to include GPS and SP messages

Tests:
- Add unit tests for SpeedLimitResolver to validate source selection, data parsing, offset application, and map data age handling

### DIFF
--- a/sunnypilot/selfdrive/controls/lib/speed_limit/speed_limit_resolver.py
+++ b/sunnypilot/selfdrive/controls/lib/speed_limit/speed_limit_resolver.py
@@ -65,9 +65,9 @@ class SpeedLimitResolver:
     if self.offset_type == OffsetType.off:
       return 0
     elif self.offset_type == OffsetType.fixed:
-      return self.offset_value * (CV.KPH_TO_MS if self.is_metric else CV.MPH_TO_MS)
+      return float(self.offset_value * (CV.KPH_TO_MS if self.is_metric else CV.MPH_TO_MS))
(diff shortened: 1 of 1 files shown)


Qubitium:
@sunnyhaibin  How about speed src of `config`? And associated ability for user to set a ceiling speedlimit.. For example, if installing a sunnypiot in a `nanny` state for a kid , rental for a friend, etc, or for whatever reason, when I would want the speed limit to be set to 120km/h or something like that. Like an always on speed limit if none already exists and if a speed limit exists, always picking the lowest value. So both map, car, and config speedlimits can coexists.

PR #1297, opened 2025-09-29
Fix params loading delay on startup

### LINKED ISSUE #1293: params value update delay
### Describe the bug

After recent rebase on top of https://github.com/sunnypilot/sunnypilot/commit/bffb2fb6fabd384a39517e46d841b45b97a58210, I noticed params became slow to load after bootup. 

```
coop_steering = get_param(CC_SP.params, "TeslaCoopSteering", False)
```
https://github.com/AmyJeanes/opendbc/blob/10483da40236f0626b125544a68f93baaabe1bbc/opendbc/sunnypilot/car/tesla/coopsteering.py#L23
In this case it takes ~35s before tesla lkas coop control kicks in (direct indication of the param switching to True). 

Image

2a251bf8a265ff32/000001bc--750a4f73a0/0

### openpilot version

52d50f984191d63cf6a79398b21df2d2e3b9236e

### Additional info

_No response_

dzid26:
Makes sure `self.param_store.update` gets processed on first call.

Partially fixes #1293

### DIFF
--- a/sunnypilot/selfdrive/controls/lib/param_store.py
+++ b/sunnypilot/selfdrive/controls/lib/param_store.py
@@ -22,7 +22,7 @@ class ParamStore:
     self.keys = universal_params + brand_params
     self._params = {}
 
-    self.frame = 0
+    self.frame = -1
 
   def update(self, params: Params) -> None:
     self.frame += 1


sunnyhaibin (approved):
Confirmed this is working. Thanks for the PR!

PR #1309, opened 2025-10-02
Device: Wake for all visual alerts

nayan8teen:
## Summary by Sourcery

Ensure the device stays awake for all visual alerts by resetting the onroad screen off timer whenever an alert is active, including e2e green light alerts.

Enhancements:
- Reset onroad sleep timer for any active visual alert rather than only non-normal alerts
- Reset onroad sleep timer for e2e green light alerts in the HUD


sunnyhaibin (approved):
LGTM

PR #1333, opened 2025-10-08
Device: Retain QuickBoot state after op switch

nayan8teen:


devtekve:
While the intention is good, the risk is high, if the branch im switching to is not prebuilt, I will be completely locked out. Perhaps do this at the end of the build process or something?

system/manager/manager.py
```diff
@@ -38,6 +39,12 @@ def manager_init() -> None:
   if params.get("DeviceBootMode") == 1:  # start in Always Offroad mode
     params.put_bool("OffroadMode", True)
 
+  # quick boot
+  if params.get_bool("QuickBootToggle") and not PC:
+    prebuilt_path = "/data/openpilot/prebuilt"
+    if not os.path.exists(prebuilt_path):
+      open(prebuilt_path, 'x').close()
```
devtekve, line 47: While the intention is good, the risk is high, if the branch im switching to is not prebuilt, I will be completely locked out. Perhaps do this at the end of the build process or something?

devtekve: Discussed offline and makes sense, there doesn't seem to be a real risk

PR #1334, opened 2025-10-08
UI: Better layout for onroad screen-off settings

nayan8teen:
image


devtekve:
i find it confusing having "brightness" and "global brightness"

sunnyhaibin:
image

PR #1337, opened 2025-10-09
[TIZI/TICI] ui: Rocket Fuel

rav4kumar:
image
image

### DIFF
--- a/selfdrive/ui/sunnypilot/onroad/rocket_fuel.py
+++ b/selfdrive/ui/sunnypilot/onroad/rocket_fuel.py
@@ -34,7 +34,6 @@ class RocketFuel:
     wp = 28.0
 
     # Draw
-    rect_w = rect.width
     rect_h = rect.height
 
     if self.vc_accel > 0:


selfdrive/ui/sunnypilot/qt/offroad/settings/visuals_panel.cc
```diff
@@ -82,6 +82,13 @@ VisualsPanel::VisualsPanel(QWidget *parent) : QWidget(parent) {
       tr("When enabled, the speedometer on the onroad screen is not displayed."),
       "",
       false,
+    },
+    {
+      "RocketFuel",
+      tr("Show Rocket Fuel Bar"),
+      tr("Show a indicator on the left side of the screen to indicate acceleration and braking."),
```
sunnyhaibin:
```suggestion
      tr("Display Rocket Fuel Bar"),
      tr("Show an indicator on the left side of the screen to display real-time vehicle acceleration and deceleration."),
```

selfdrive/ui/sunnypilot/qt/onroad/model.cc
```diff
@@ -25,6 +25,7 @@ void ModelRendererSP::drawPath(QPainter &painter, const cereal::ModelDataV2::Rea
   auto *s = uiState();
   auto &sm = *(s->sm);
   bool blindspot = Params().getBool("BlindSpot");
+  bool rocketfuel = Params().getBool("RocketFuel");
```
sunnyhaibin: These should be done inside ui.cc with `UIScene` to avoid frequent calls with Params: https://github.com/sunnypilot/sunnypilot/blob/3df1b53fab4341074ace53b2797eb16eb5ed3e09/selfdrive/ui/sunnypilot/ui.cc#L61-L75

selfdrive/ui/sunnypilot/qt/onroad/model.cc
```diff
@@ -86,5 +87,30 @@ void ModelRendererSP::drawPath(QPainter &painter, const cereal::ModelDataV2::Rea
   } else {
     // Normal path rendering
     ModelRenderer::drawPath(painter, model, surface_rect.height());
+
+    if (rocketfuel) {
+      rocketFuel(painter, surface_rect);
+    }
+
+
```
sunnyhaibin: Why do we only draw this when not rainbow path?

selfdrive/ui/sunnypilot/qt/onroad/model.cc
```diff
@@ -86,5 +87,30 @@ void ModelRendererSP::drawPath(QPainter &painter, const cereal::ModelDataV2::Rea
...
   }
 }
+
+void ModelRendererSP::rocketFuel(QPainter &p, const QRect &surface_rect) {
+  UIState *s = uiState();
+  float accel = (*s->sm)["carControl"].getCarControl().getActuators().getAccel();
```
sunnyhaibin: These should be done in `ModelRendererSP::draw` (does not exist, but should be done there) so we avoid recreating `uiState()`.

sunnyhaibin:
- Rocket Fuel is meant to display the forward acceleration of the car, instead of the long actuator's acceleration output?
- Draws too far which goes behind the bottom Dev UI.
image

devtekve (changes requested):
Since this feature on it's current state on this PR has deviated from what we had in mind (and from what we had on previous releases), we will spend more time on this to ensure the experience is consistent to what the initial expectation was. This means we'd probably need to switch from using `carControl.actuators.accel` which is only available when OP long is enabled and commanding to `carState.aEgo` which is available regardless. But we will address this post-release as it's high-risk low-reward

PR #1343, opened 2025-10-09
ICBM: always make sure button timer is running

sunnyhaibin:
## Summary by Sourcery

Ensure the button timer and speed limit assist logic always run correctly under non-PCM cruise control by unifying the enable check and simplifying conditions

Enhancements:
- Change update_speed_limit_assist_v_cruise_non_pcm to take an enabled flag and early-exit when disabled
- Remove redundant enabled check in the non-PCM cruise conditional to always invoke speed limit assist and button timer


sunnyhaibin:
This fix isn't it. Closing for now.

PR #1353, opened 2025-10-10
[TICI] Sync: `master` into `master-tici`

sunnyhaibin:


mouxangithub:
I downloaded and installed it on c3 and it was successfully installed, but the Black Panda seems to have been unable to connect all along

Issue #1217, opened 2025-09-05
TICI: limited support checklist


sunnyhaibin:
## Branching Strategy
- Current gen
  - Main trunk:
    - [x] `master`
    - [x] `master-dev-c3-new` > `master-dev`
  - Prebuilts
    - [x] `staging-c3-new` > `staging`
    - [x] `dev-c3-new` > `dev`
    - [ ] `release-c3` > `release-tizi` (at next release cut)
- TICI
  - Main trunk:
    - [x] `master-tici`
  - Prebuilts
    - [x] `staging-tici`
    - [ ] `release-tici` (at next release cut)
  - Notes
    - No `dev` related main trunk or prebuilts

## Sync/Merge Strategy
- sunnypilot/sunnypilot
  - `master` < Upstream
  - `master-tici` < `master` (sunnypilot-specific changes only)
- sunnypilot/opendbc
  - `master` < Upstream
  - `master-tici` < `master` (sunnypilot-specific changes only)
    - Car ports and safety changes should be:
      - Merged into `master` first
      - Then optionally merged into `master-tici`
- sunnypilot/panda
  - `master` < Upstream
  - `master-tici` < `master` (sunnypilot-specific changes only)

## Implementation Checklist
- [x] Migrate users to applicable branches if TICI: https://github.com/sunnypilot/sunnypilot/pull/1212
- [x] CI updated for new branch names: https://github.com/sunnypilot/sunnypilot/pull/1225
- [x] Prebuilt pipelines updated (`tici` related changes must be in `master-tici` only): https://github.com/sunnypilot/sunnypilot/pull/1220 
- [ ] Auto PRs CI for `master-tici` < `master` (sunnypilot-specific changes only)
- [x] Branch protection rules applied
- [x] Documentation updated
- [x] Old branch refs cleaned up in workflows and other components: https://github.com/sunnypilot/sunnypilot/pull/1225

##  Notes
- sunnypilot-specific changes:
  - `master-tici` is derived from `master` as of https://github.com/sunnypilot/sunnypilot/commit/b161764b1e9a4152c9641b3a4168c7e88960e59e, but not all changes in `master` are guaranteed to flow into `master-tici`.
  - Whether a feature, car port, or update is merged into `master-tici` is at the full discretion of the sunnypilot team.
  - `master-tici` will not be maintained as tightly or consistently as `master`.
- Upstream syncs:
  - Upstream changes flow into `master`.
  - These do not automatically flow into `master-tici`.
[...]

PR #1241, opened 2025-09-12
Docs - Update README.md with updated branch names

isaacdontjelindell:
Update README to reference correct branch names for staging and dev, per this thread: https://discord.com/channels/880416502577266699/994915271213064222/1416106276802400451

## Summary by Sourcery

Documentation:
- Remove old '-c3-new' suffix and update branch names for staging, dev, and release in the installation URL table

PR #1354, opened 2025-10-10
docs: fix README.md install instructions referring to orphaned branches

borhub:
Align with https://github.com/sunnypilot/sunnypilot/issues/1217 and with how the installer works these days on 3X devices.

Also remove section referring to even older legacy branches.

The main reason for the PR is so that a fresh user can easily switch from OP to SP by reading the SP's README.md.
Before this change, the user is met with software download errors/404s when following the old README with the old branch names/URLs.

### DIFF
--- a/README.md
+++ b/README.md
@@ -43,11 +43,11 @@ Please refer to [Recommended Branches](#-recommended-branches) to find your pref
 
 |      Branch      |                 Installation URL              |
 |:----------------:|:---------------------------------------------:|
-| `staging` | `https://staging.sunnypilot.ai`        |
-|   `dev`   | `https://dev.sunnypilot.ai`            |
+| `staging` | `https://install.sunnypilot.ai/staging`        |
+|   `dev`   | `https://install.sunnypilot.ai/dev`            |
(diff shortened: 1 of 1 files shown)


borhub:
Looking at the file history, the `#-recommended-branches` link was never set up so I didn't mess with that and left as is.

borhub:
Superceded by https://github.com/sunnypilot/sunnypilot/pull/1453

@devtekve @sunnyhaibin  Looks like no one reads PRs by default and core devs need to be pinged in Discord or Discourse to get things looked at?

devtekve:
> Superceded by #1453
> 
> @devtekve @sunnyhaibin Looks like no one reads PRs by default and core devs need to be pinged in Discord or Discourse to get things looked at?

Hi, sorry about that! But yeah, pinging us helps, you can also try pinging us here on GitHub but you'll get better followup if you ping us on the community forum, that's what we aim to monitor more often.. In this case we also had this one https://github.com/sunnypilot/sunnypilot/pull/1241 which had a request for changes https://github.com/sunnypilot/sunnypilot/pull/1241#issuecomment-3447966753 and I genuinely thought that was the only one 😅

PR #1368, opened 2025-10-13
UI: Update Traditional Chinese translation

dsa302010:


sunnyhaibin:
Thanks for the PR. Can you revert the translations that were already previously translated?

PR #1385, opened 2025-10-15
interfaces: clean up unsupported params during initialization

sunnyhaibin:
## Summary by Sourcery

Clean up unsupported parameters during interface setup by removing irrelevant tuning and feature flags based on the vehicle’s steering and longitudinal control configuration, and set a default speed limit mode when needed.

Enhancements:
- Introduce _cleanup_unsupported_params to remove unsupported neural network lateral control, torque control, cruise management, and custom ACC parameters based on CarParams settings
- Invoke the cleanup routine in setup_interfaces to enforce parameter validity and default SpeedLimitMode.warning where appropriate

### DIFF
--- a/selfdrive/ui/sunnypilot/qt/offroad/settings/longitudinal/speed_limit/speed_limit_settings.cc
+++ b/selfdrive/ui/sunnypilot/qt/offroad/settings/longitudinal/speed_limit/speed_limit_settings.cc
@@ -123,7 +123,7 @@ void SpeedLimitSettings::refresh() {
     has_longitudinal_control = hasLongitudinalControl(CP);
     intelligent_cruise_button_management_available = CP_SP.getIntelligentCruiseButtonManagementAvailable();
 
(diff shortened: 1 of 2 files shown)


sunnypilot/selfdrive/car/interfaces.py
```diff
@@ -66,6 +67,24 @@ def _initialize_torque_lateral_control(CI: CarInterfaceBase, CP: structs.CarPara
     CI.configure_torque_tune(CP.carFingerprint, CP.lateralTuning)
 
 
+def _cleanup_unsupported_params(CP: structs.CarParams, CP_SP: structs.CarParamsSP, params: Params = None) -> None:
```
devtekve, line 70: Can we carlog as warnings when we are clearing up certain things to leave a breadcrumb for us ? I feel this may be forgotten at some point and we'd be wondering why the heck we see a particular behavior

PR #1422, opened 2025-10-23
ui: openpilot Longitudinal Control → sunnypilot Longitudinal Control

sirmuskrat:
Changing mentions in the openpilot Longitudinal Control ui option to say sunnypilot instead. This aligns with the Experimental Mode ui option telling users "sunnypilot Longitudinal Control" must be enabled for Experimental mode. 

This also more accurately describes the feature as Sunnypilot adds modifications and changes to the way Longitudinal Control (Alpha) functions.

## Summary by Sourcery

Update the developer panel UI to rebrand the alpha longitudinal control option from openpilot to sunnypilot, ensuring labels, warnings, and descriptions reflect the sunnypilot branding.

Enhancements:
- Rebrand "Longitudinal Control (Alpha)" option label to use sunnypilot instead of openpilot
- Update warning text to mention sunnypilot longitudinal control disables AEB
- Adjust description text to default to built-in ACC and switch to sunnypilot longitudinal control


sunnyhaibin:
When we move onto Raylib, all stock toggles and descriptions are untouched, so mentions of the old name would still need to be updated. We won't be modifying them until we have migrated over from Qt to Raylib.

Opened an issue to track this:
- https://github.com/sunnypilot/sunnypilot/issues/1423

PR #1449, opened 2025-11-02
Update joystick axis mappings for proper control

Sliicy:
This is the corrected output for Xbox One controller to control gas/brakes and steering using both joysticks on the gamepad.

I actually made this change upstream a few years ago but it was overwritten for some reason...

## Summary by Sourcery

Update joystick axis mappings for proper control on Xbox One controllers

Enhancements:
- Use ABS_Y for acceleration/brake and ABS_RX for steering in non-longcontrol mode
- Update flip_map to map ABS_RY to the new acceleration axis


devtekve:
Haven't checked the context of the pr, change seems simple, but does this break playstation's controller?

Sliicy:
Hi, I don't have a PlayStation controller to test with. But I know that with the Xbox One controller that I tested with last night, this change seems to be the correct button layout.
It would be awesome if we can implement a feature where it detects what kind of controller is connected as well; this way if the PlayStation uses a different layout it would work also.

PR #1455, opened 2025-11-04
Update ui to observe the IsMetric param

fl0yd:
## Summary by Sourcery

Fix the handling of the IsMetric parameter in the sunnypilot UI to ensure lead car distance updates immediately when toggling between metric and imperial units without requiring a UI restart.

Bug Fixes:
- Watch the "IsMetric" parameter in the Sunnypilot UI's parameter watcher to detect unit setting changes.
- Refresh base OpenPilot parameters (including is_metric) on startup and when parameters change by calling ui_update_params in ui_update_params_sp.

Enhancements:
- Initialize Sunnypilot-specific parameters only after base UI parameters are loaded to guarantee correct unit conversion.

### DIFF
--- a/selfdrive/modeld/SConscript
+++ b/selfdrive/modeld/SConscript
@@ -51,8 +51,8 @@ def tg_compile(flags, model_name):
 for model_name in ['driving_vision', 'driving_policy', 'dmonitoring_model']:
   flags = {
     'larch64': 'DEV=QCOM',
-    'Darwin': 'DEV=CPU IMAGE=0',
-  }.get(arch, 'DEV=LLVM IMAGE=0')
+    'Darwin': f'DEV=CPU HOME={os.path.expanduser("~")} IMAGE=0', # tinygrad calls brew which needs a $HOME in the env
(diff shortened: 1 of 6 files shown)


selfdrive/ui/sunnypilot/ui.cc
```diff
@@ -32,6 +32,9 @@ UIStateSP::UIStateSP(QObject *parent) : UIState(parent) {
     "carStateSP", "liveParameters", "liveMapDataSP", "carParamsSP"
   });
 
+  // Initialize parameters including is_metric
+  ui_update_params_sp(this);
```
devtekve: im not sure this is needed, the.`StandstillTimer` or the `DevUIInfo` haven't needed it though

fl0yd: Upon further review, I agree. It will get picked up when ui_update_params_sp() runs. I'll fix the commit to remove this portion.

fl0yd:
I have updated the commit to strike the unneeded constructor

sunnyhaibin:
Thanks for the PR. sunnypilot is now using Raylib as its UI framework; therefore this PR is no longer applicable.

PR #1462, opened 2025-11-09
UI: Russian translation

Falseclock:
just translation, nothing else


sunnyhaibin:
We're now using Raylib and no longer using Qt. Wanna open a new PR with the translations on Raylib? Thanks!

PR #1466, opened 2025-11-11
Display: Onroad Screen-Off - prevent wake-up for normal alerts

nayan8teen:


sunnyhaibin:
Raylib has taken over.

PR #1475, opened 2025-11-15
ui: sunnypilot toggle style

nayan8teen:
image

### DIFF
--- a/system/ui/widgets/scroller.py
+++ b/system/ui/widgets/scroller.py
@@ -28,8 +28,8 @@ class LineSeparator(Widget):
     self._rect.width = parent_rect.width
 
   def _render(self, _):
-    rl.draw_line(int(self._rect.x), int(self._rect.y),
-                 int(self._rect.x + self._rect.width), int(self._rect.y),
+    rl.draw_line(int(self._rect.x) + LINE_PADDING, int(self._rect.y),
+                 int(self._rect.x + self._rect.width) - LINE_PADDING, int(self._rect.y),
                  LINE_COLOR)
 
 


TheSecurityDev:
What's the reason exactly for making our own toggle component instead of reusing the one from openpilot and just changing the color or something? I get we'll have to make custom components for a lot of things, but it would be nice to reuse where possible.

nayan8teen:
> What's the reason exactly for making our own toggle component instead of reusing the one from openpilot and just changing the color or something? I get we'll have to make custom components for a lot of things, but it would be nice to reuse where possible.

It's more than just a different color. It updates the toggle design, renders it on the left instead of right to align with the overall sp design language, and adds an optional param to make it easy to use - don't need to handle a callback to set/unset param every time a toggle is used.

Also, this was the very first poc for us to see how we would go about extending the new op raylib code without making future syncs any more complicated.

system/ui/sunnypilot/widgets/toggle.py
```diff
@@ -0,0 +1,100 @@
...
+    # Draw outline by drawing a slightly larger rounded rectangle behind the background
+    outline_rect = rl.Rectangle(bg_rect.x - 2, bg_rect.y - 2, bg_rect.width + 4, bg_rect.height + 4)
+    rl.draw_rectangle_rounded(outline_rect, 1.0, 10, outline_color)
+
+    # Draw actual background
+    rl.draw_rectangle_rounded(bg_rect, 1.0, 10, bg_color)
+
+    # Draw knob to sit inside the background
+    knob_padding = 5
+    knob_radius = style.TOGGLE_BG_HEIGHT / 2 - knob_padding
```
TheSecurityDev: I think these could both be constants outside the render function.

TheSecurityDev:
> It's more than just a different color. It updates the toggle design, renders it on the left instead of right to align with the overall sp design language, and adds an optional param to make it easy to use - don't need to handle a callback to set/unset param every time a toggle is used.
> 
> Also, this was the very first poc for us to see how we would go about extending the new op raylib code without making future syncs any more complicated.

Sure that's fine, I guess I'm just wondering why it's important to have a different toggle design for sunnypilot to begin with. I know it's different with the current UI and I've always wondered the reason. Also, if you come across any improvements to the existing widgets, please open a PR on openpilot. I know it's easier to sync when the code is separated but I think it's nice to share as much as we can.

system/ui/sunnypilot/widgets/toggle.py
nayan8teen: Agreed, good point!

nayan8teen:
> > It's more than just a different color. It updates the toggle design, renders it on the left instead of right to align with the overall sp design language, and adds an optional param to make it easy to use - don't need to handle a callback to set/unset param every time a toggle is used.
> > 
> > Also, this was the very first poc for us to see how we would go about extending the new op raylib code without making future syncs any more complicated.
> 
> Sure, I guess I'm just wondering why it's important to have a different toggle design for sunnypilot to begin with. I know it's different with the current UI and I've always wondered the reason. Also, if you come across any improvements to the existing widgets, please open a PR on openpilot. I know it's easier to sync when the code is separated but I think it's nice to share as much as we can.

That's the beauty of forks though, is it not? Yes, the improvements and additions to selfdriving are a big part, but so is the UI. It gives the forks their own personality.

In terms of doing a PR upstream, we can and do absolutely do that. But we would also not wait for something small like this to be accepted and merged upstream instead of just doing it on our fork.

TheSecurityDev:
> That's the beauty of forks though, is it not? Yes, the improvements and additions to selfdriving are a big part, but so is the UI. It gives the forks their own personality.
> 
> In terms of doing a PR upstream, we can and do absolutely do that. But we would also not wait for something small like this to be accepted and merged upstream instead of just doing it on our fork.

Sure, makes sense :) I just don't want you to have to waste time reimplementing stuff that's already been done. Looks like you've made some good progress though, I can't wait to finally use raylib on my device.

devtekve:
> ## raylib UI Preview
> All Screenshots

Fuck yeah! Preview is back

nayan8teen:
> > ## raylib UI Preview
> > All Screenshots
> 
> Fuck yeah! Preview is back

It's stock though. Just pushed an update to have sp ui previews 🤘

TheSecurityDev:
> It's stock though. Just pushed an update to have sp ui previews 🤘

The issue with the confirmation dialog screenshot is due to the different toggle position. We'll need a separate screenshot script to account for all the sunnypilot differences in the future.

For now though it's cool to see the comparison between stock and this update.

selfdrive/ui/tests/test_ui/raylib_screenshots.py
```diff
@@ -292,6 +292,7 @@ def create_screenshots():
     with OpenpilotPrefix():
       params = Params()
       params.put("DongleId", "123456789012345")
+      params.put_bool("sunnypilot_ui", True)
```
TheSecurityDev: I would create a separate `raylib_screenshots_sp.py` file for this so we can handle all the sunnypilot differences, and update the PR workflow accordingly (or add a separate one and leave the stock report alone).

TheSecurityDev: Oh and do in a separate PR, although it's nice to see it working on this one.

TheSecurityDev: Or alternatively we could use the SP UI flag to determine which click positions to use and which test cases to run (or just overwrite if we don't want the stock ones), which will allow reusing the code but make future merges slightly more complicated. It might be worth the trade off though.

nayan8teen:
> > It's stock though. Just pushed an update to have sp ui previews 🤘
> 
> The issue with the confirmation dialog screenshot is due to the different toggle position. We'll need a separate screenshot script to account for all the sunnypilot differences in the future.
> 
> For now though it's cool to see the comparison between stock and this update.

Yeah, will be fixing that. The images are also not right for some reason. Even though it picked up the diff as expected, the images in the report are all from master lol.


EDIT - WTH - the images are fine now. Without any changes lol

selfdrive/ui/tests/test_ui/raylib_screenshots.py
TheSecurityDev: Ok ok, after thinking about it a lot I'm thinking your approach might be the best (and simplest) one after all. Just modify this one file, and any merge conflicts will let us know when the script needs updated. I have some open PRs for stock that would be nice to have here too, so this way we'll have all the same code for the most part.

If it becomes too big to manage, we can break into a separate class or something, or a new script if necessary. But no need to get ahead of things.

PR #1487, opened 2025-11-18
[TIZI/TICI] ui: chevron metrics

rav4kumar:
image


devtekve:
we probably should try to move this to it's own sunnypilot folder like we used to on qt

sunnyhaibin:
## After decoupling (no change)

image

sunnyhaibin:
## Final

image

PR #1488, opened 2025-11-18
ui: sunnylink client-side implementation

nayan8teen:
Migrate the sunnylink client from cpp to python for raylib ui integration.

### DIFF
--- a/sunnypilot/sunnylink/sunnylink_state.py
+++ b/sunnypilot/sunnylink/sunnylink_state.py
@@ -1,3 +1,9 @@
+"""
+Copyright (c) 2021-, Haibin Wen, sunnypilot, and a number of other contributors.
+
+This file is part of sunnypilot and is licensed under the MIT License.
+See the LICENSE.md file in the root directory for more details.
+"""
 from enum import IntEnum
 import threading
 import time


common/params_keys.h
```diff
@@ -173,6 +173,7 @@ inline static std::unordered_map<std::string, ParamKeyAttributes> keys = {
     {"ShowAdvancedControls", {PERSISTENT | BACKUP, BOOL, "0"}},
     {"ShowTurnSignals", {PERSISTENT | BACKUP, BOOL, "0"}},
     {"StandstillTimer", {PERSISTENT | BACKUP, BOOL, "0"}},
+    {"sunnypilot_ui", {PERSISTENT, BOOL, "1"}},
```
devtekve: I have a feeling this should be the default (aka not needing a param, and only having a param if we want stock ui instead)

devtekve: Maybe it’s just pride 🤣 this should work regardless

sunnypilot/sunnylink/sunnylink_state.py
```diff
@@ -0,0 +1,188 @@
...
+      response = self._api.api_get(f"device/{self.sunnylink_dongle_id}/users", method='GET', access_token=token)
+      if response.status_code == 200:
+        users = response.text
+        self._params.put("SunnylinkCache_Users", users)
+        with self._lock:
+          _parse_users(users)
+    except Exception as e:
+      cloudlog.exception(f"Failed to fetch sunnylink users: {e} for dongle id {self.sunnylink_dongle_id}")
+
+  def _worker_thread(self) -> None:
```
devtekve, line 162: Can we gate the fetches to only happen if as have internet connectivity ?

common/params_keys.h
nayan8teen: Lol, it will work the same way regardless. 🤣

sunnypilot/sunnylink/sunnylink_state.py
nayan8teen: Done

PR #1489, opened 2025-11-19
ui: `UIStateSP`

nayan8teen:


devtekve:
this feels inverted, we'd normally inherit from `UIState` and use `UIStateSP` on our stuff and beyond

nayan8teen:
> this feels inverted, we'd normally inherit from `UIState` and use `UIStateSP` on our stuff and beyond

That is exactly where we started. The issue with thay is ui_state is imported everywhere. And when I didn't change them to ui_state_sp with that approach, it was causing weird behavior. 

This felt better as we are still able to keep our changes isolated, but no need to change any imports anywhere.

devtekve:
> > this feels inverted, we'd normally inherit from `UIState` and use `UIStateSP` on our stuff and beyond
> 
> That is exactly where we started. The issue with thay is ui_state is imported everywhere. And when I didn't change them to ui_state_sp with that approach, it was causing weird behavior.
> 
> This felt better as we are still able to keep our changes isolated, but no need to change any imports anywhere.

Okay I don't love it but if it works that's good for this early stage IMHO. I will let @sunnyhaibin call the final shot on this one, but I won't let perfect be the enemy of progress

sunnyhaibin:
I agree with @nayan8teen - the original design seems to be the safer and more predictable option right now. Keeping `UIState` as the concrete class and layering our additions through `UIStateSP` avoids fighting the singleton, avoids breaking every `ui_state` import, and keep things cleaner should upstream decides to change this drastically down the road.

We’ll stick with this approach for now and revisit the structure later if it becomes a limitation.

sunnyhaibin (approved):

PR #1502, opened 2025-11-21
ui: `SimpleButtonActionSP`

nayan8teen:
sample:
image


devtekve:
I like it but the styling looks very off from what we used to know and love, can we maybe make the button a bit bigger? Maybe it just needs sync up since this one is quite old

devtekve:
Syncd this one too, tho i dont know if i did it right :D i only added the simple button to list_view

nayan8teen:
> I like it but the styling looks very off from what we used to know and love, can we maybe make the button a bit bigger? Maybe it just needs sync up since this one is quite old

yeah, not sure why I had that old screenshot here. Updated to the current look. And here's how it looks on the lateral panel PR:

image

sunnyhaibin (dismissed):
LGTM

sunnyhaibin (approved):

PR #1504, opened 2025-11-22
[TIZI/TICI] ui: MICI style turn signals

rav4kumar:
mici_right
mici_left


devtekve:
I think we should try to do our customizations on the sunnypilot folder for the ui stuff

PR #1505, opened 2025-11-22
navd: destination autofill

Discountchubbs:
nav auto fill while typing destination... it does look a little ugly so changes will be made ofc but here it is


Discountchubbs:
This will be fabulous on SL

PR #1509, opened 2025-11-23
ui: vehicle brand settings

Discountchubbs:


devtekve:
Custom longitudinal tuning shouldn't be visible for vehicles that don't support long control, it is currently visible but disabled

sunnyhaibin (approved):
Full send. No _ragrets_.

PR #1512, opened 2025-11-25
ui: display mouse coordinates with debug mode

sunnyhaibin:


TheSecurityDev:
Do you think you could get this upstream? It would be super useful for me there.

PR #1513, opened 2025-11-25
ui: reduce sidebar `scroll` in UI Preview

sunnyhaibin:


TheSecurityDev:
I have a PR #1519  that I think should fix what it seems you were trying to.

PR #1521, opened 2025-11-27
ui: Developer panel extension

Discountchubbs:
Relies on first to do the python execution of external_storage.py, but doesnt depend on it for merging. you will see i have it ready for when the code is added, but it is simply commented out until then :)


royjr:
see https://github.com/sunnypilot/sunnypilot/pull/1524 for external storage

sunnyhaibin:
This needs to not be center aligned haha
image

Edit: Split in https://github.com/sunnypilot/sunnypilot/pull/1521

### New

image

PR #1524, opened 2025-11-28
ui: external storage

royjr:


royjr:
should probably kill this feature, not used by many

PR #1539, opened 2025-12-01
ui: lateral-only and longitudinal-only UI statuses support

royjr:
- [x] enable certain elements when mads only
- [x] dont override with manual steering on long only
- [x] move to sp


devtekve:
Reminder to try to scope the code as much as humanly possible on sunnypilot folder

PR #1549, opened 2025-12-03
[UPSTREAM] ui: better experimental alert

royjr:
with the smaller mici screen, it might be harder to tell whether you're in chill or experimental mode


greatgitsby:
this feature is planned in upstream (i was helping with a first impl).

feel free to add but know it’s coming

sunnyhaibin:
@royjr lmk when it's rebased, we can merge just the event/alert for now.

PR #1559, opened 2025-12-05
ui: toggle should only toggle toggle - don't toggle description with toggle. Yeah, that!

nayan8teen:
Before:

https://github.com/user-attachments/assets/8acb1740-87bc-41d0-aad1-8beb289d2605

After:

https://github.com/user-attachments/assets/92a30300-6e6d-411a-a924-0fb6d858197e


sunnyhaibin (approved):
image

PR #1576, opened 2025-12-15
ci: use Brewfile for macOS setup and update Homebrew cache keys

sunnyhaibin:


sunnyhaibin:
Let's give it a try again

PR #1593, opened 2025-12-20
sunnylink: block more params

dzid26:
**Description**

Inspired by https://github.com/sunnypilot/sunnypilot/pull/1591 I extended remote parameter protection to block additional security-sensitive parameters. This has minor impact on backup/restore convenience in exchange for security hardening and privacy piece of mind.

Additionally I blocked any remote modifications while OP is engaged for safety reasons. - If needed, some low impact params can be whitelisted in the future.
Todo: toggle state gets out of sync if modified while driving, so that's something that could be improved on the front end side. 

**Verification**

I tested safety implication of pushing toggles while driving using https://www.sunnylink.ai/dashboard.  

- Before the change pushing Offroad On toggle would disengage suddenly and play siren (after a delay and for a fraction of a second - probably panda bug), not giving the driver much time to react. 

- After the change, pushing Offroad On toggle didn't have any effect - as expected.

### DIFF
--- a/sunnypilot/sunnylink/athena/sunnylinkd.py
+++ b/sunnypilot/sunnylink/athena/sunnylinkd.py
@@ -42,11 +42,45 @@ METADATA_PATH = os.path.join(os.path.dirname(os.path.dirname(os.path.abspath(__f
 
 params = Params()
 
-# Parameters that should never be remotely modified for security reasons
-BLOCKED_PARAMS = {
-  "GithubUsername",  # Could grant SSH access
-  "GithubSshKeys",   # Direct SSH key injection
+# Parameters that should never be remotely modified
+ALWAYS_BLOCKED_PARAMS = {
+  "GithubSshKeys",
+  "GithubUsername",
+  "SunnylinkAllowSensitiveWrite",
 }
+
+# Parameters blocked by default, but can be unlocked via SunnylinkAllowSensitiveWrite toggle
+SENSITIVE_PARAMS = {
+  # SSH/Authentication
+  "SshEnabled",
+  "AdbEnabled",
+  "EnableCopyparty",
+  "GsmApn",
+
+  # Remote code execution
+  "EnableGithubRunner",
+  "DisableUpdates",
+
+  # Require physical presence
+  "LongitudinalManeuverMode",
+  "JoystickDebugMode",
+
+  # Privacy
+  "RecordFront",
+  "RecordAudio",
+  "RecordAudioFeedback",
+  "EnableSunnylinkUploader",
+
+  # Identity
+  "SunnylinkDongleId",
+  "DongleId",
+  "AccessToken",
+
+  # Legal
+  "HasAcceptedTerms",
+  "CompletedTrainingVersion",
+}
+
 
 
 def handle_long_poll(ws: WebSocket, exit_event: threading.Event | None) -> None:
@@ -254,11 +288,16 @@ def getParams(params_keys: list[str], compression: bool = False) -> str | dict[s
 @dispatcher.add_method
 def saveParams(params_to_update: dict[str, str], compression: bool = False) -> None:
   is_engaged = params.get_bool("IsEngaged")
+  allow_sensitive = params.get_bool("SunnylinkAllowSensitiveWrite")
 
   for key, value in params_to_update.items():
-    # disallow modifications to blocked parameters
-    if key in BLOCKED_PARAMS:
-      cloudlog.warning(f"sunnylinkd.saveParams.blocked: Attempted to modify blocked parameter '{key}'")
+    if key in ALWAYS_BLOCKED_PARAMS:
+      cloudlog.warning(f"sunnylinkd.saveParams.always_blocked: Attempted to modify '{key}'")
+      continue
+
+    # Block sensitive params unless toggle is enabled
+    if key in SENSITIVE_PARAMS and not allow_sensitive:
+      cloudlog.warning(f"sunnylinkd.saveParams.sensitive_blocked: Attempted to modify '{key}'")
       continue
 
     # Block all params while engaged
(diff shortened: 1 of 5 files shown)


sunnypilot/sunnylink/athena/sunnylinkd.py
```diff
@@ -42,13 +42,47 @@
...
+  "EnableSunnylinkUploader",
+
+  # Identity
+  "SunnylinkDongleId",
+  "DongleId",
+  "AccessToken",
+
+  # Legal
+  "HasAcceptedTerms",
+  "CompletedTrainingVersion",
```
devtekve, line 81: All of this prevents an uneventful migration from devices (we support device migration for people going from c3/x to c4), so before accepting this it'd be great if there's also a flow added for migration cases

cc @sunnypilot/dev-internal

dzid26: I was thinking the Restore function could temporarily allow for writing the backup params.  
Also the backed up `GithubUsername` and `GithubSshKeys` are already blocked from being restored.

devtekve: The restore function will not be used because the backups are encrypted with different keys, and restore shall be unrestricted from this

Discountchubbs: Disable updates param has nothing to do with remote code execution. It is a stock openpilot param, not one we made. If you have questions about that, I encourage you to look at updater and updated.

Discountchubbs: Give us a couple hours to do a formal research write up and get back to you @dzid26 

In the meantime, enjoy your day!

dzid26: > Disable updates param has nothing to do with remote code execution

It's about giving the user a piece of mind in my view. 

In theory updates allow for pushing arbitrary code, which in turn can allow remote execution.
 
It's stock, but I believe comma api cannot change it. SL can. 

In this particular case it would be best to not allow to change it to False. :)

Discountchubbs:
Thanks for the work on this. After reviewing the latest changes and comments, we’re not going to move forward with this PR **in its current form**.

A few clarifications, since some assumptions here don’t align with how the system actually works:

## Key Concerns

### Updates and "RCE"

>It's about giving the user a piece of mind in my view.
>
>In theory updates allow for pushing arbitrary code, which in turn can allow remote execution.
>
>It's stock, but I believe comma api cannot change it. SL can.
>
>In this particular case it would be best to not allow to change it to False. :)

`DisableUpdates` is a stock openpilot parameter and is not a remote code execution mechanism. Updates are delivered through the standard updater flow and applied on reboot[^1][^2]. During restarts, sunnylink does not have an active control plane connection. Framing this parameter as an RCE vector is not accurate and does not reflect the actual threat model.

More broadly, sunnylink does not introduce any new code-delivery mechanism beyond what already exists in upstream openpilot. If a user fundamentally distrusts the update model itself, that concern applies equally to openpilot / sunnypilot as a whole, not to sunnylink specifically.

### Migration
As noted by `@devtekve`, the "Three-Layer Confirmation" blocks seamless comma-to-comma device migration. Backups and restores rely on these parameters. Blocking them forces users into manual configuration, which lowers reliability of sunnylink backup and restore feature we provide. 

### Misclassification of Risks, Scope, and Threat Model
This PR attempts to mitigate a hypothetical adversarial-operator scenario by selectively blocking parameters. That framing is out of scope for sunnylink’s design.

sunnylink operates on a **server-trust / chain-of-trust** model[^3], which we’ve mentioned multiple times[^4][^5]. If that model doesn’t work for a user, the supported options are to **disable sunnylink entirely**, or to **fork sunnypilot and modify `sunnylinkd.py`** to enforce stricter device-side controls. Changing defaults or selectively blocking parameters does not meaningfully alter that trust boundary.

Re-introducing variations of this approach does not move us closer to something we’re willing to ship.

For additional context from an independent community member who reviewed this exact class of claims and the sunnylink trust model in detail, you may find the following analysis useful:
https://gist.github.com/zikeji/c0cfdd9144a8a680ced3e17855e76fb1

## Path Forward
We are open to a much smaller, targeted PR(s) if you wish to split this up.
* **Potential candidates for removing from backup:** `SshEnabled` or `RecordFront` (Privacy focused).
* **Out of scope:** `DisableUpdates`, `AccessToken`, and the complex confirmation dialogs on the UI. 

## The Challenge
If you believe the current API has a critical vulnerability that this blacklist solves, we invite you to demonstrate it. Treat yourself as a "bad actor" and attempt to compromise your device  via the sunnylink API, and open a PR or issue to support your claim. 

---

### References

1. **OWASP A01:2021 – Broken Access Control**
   https://owasp.org/Top10/A01_2021-Broken_Access_Control/
   *Principle: Focus on verifying the user, not just filtering the object.*

2. **The Protection of Information in Computer Systems (Saltzer & Schroeder)**
   https://web.mit.edu/Saltzer/www/publications/protection/
   *Principle: "Psychological Acceptability" — Security must not unduly interfere with user workflow.*

[^1]: https://github.com/sunnypilot/sunnypilot/blob/34ce7468695dac0469f8f9ece4bd9e03d2ce8b9b/system/updated/updated.py
[^2]: https://github.com/commaai/openpilot/blob/a8cfa2e2feaf63fde0f34d6e036c0ad5498d02eb/system/updated/updated.py
[^3]: https://community.sunnypilot.ai/t/sunnylink-to-comma-request-authentication/1857/6?u=sunnyhaibin
[^4]: https://github.com/sunnypilot/sunnypilot/pull/1605#issuecomment-3682098271
[^5]: https://github.com/sunnypilot/sunnypilot/pull/1609#issuecomment-3683987641

dzid26:
> That framing is out of scope for sunnylink’s design.
 Sunnylink operates on a server-trust / chain-of-trust model[3](https://github.com/sunnypilot/sunnypilot/pull/1593#user-content-fn-3-38287b1e8a66679f0898f77902674ddc)

`GithubUsername` and `GithubSshKey` got recently blocked from restoring from backup and remote changes. This signaled that you don’t fully trust your server-trust/chain-of-trust model for _certain_ actions.


> As noted by @devtekve, the "Three-Layer Confirmation" blocks seamless comma-to-comma device migration. Backups and restores rely on these parameters. Blocking them forces users into manual configuration, which lowers reliability of sunnylink backup and restore feature we provide.

https://github.com/sunnypilot/sunnypilot/blob/a04a5b4284066f3f05feede209f7ec544a79d6f7/common/params_keys.h#L49-L50

What is the plan for them? Reverting the access back or removing from the backup?

PR #1604, opened 2025-12-21
[UPSTREAM] ui: dm offset

royjr:
Screenshot_2025-12-18_at_00 21 51

works great!
- [ ] confirm rhd
- [ ] why 20?
- [ ] dynamic from position?


sunnyhaibin:
This is now updated in upstream with similar behaviors.

PR #1605, opened 2025-12-21
Sunnylink: opt-in model

dzid26:
When a feature expose private data to third parties, users should give consent before the data leaves the device. 

This PR adds:
- "opt-in" by default
- consent dialog explaining that sunnylink admins can access location, settings, and GitHub identity when paired.
- sneaked-in dongle id display fix for the PC

We can work on wording if desired.

PR #1609, opened 2025-12-22
Sunnylink: opt-out by default

dzid26:
A minimum version of https://github.com/sunnypilot/sunnypilot/pull/1605

- Gives user a chance to opt-out from Sunnylink. 

image

image

image

### DIFF
--- a/common/params_keys.h
+++ b/common/params_keys.h
@@ -201,7 +201,7 @@ inline static std::unordered_map<std::string, ParamKeyAttributes> keys = {
     {"SunnylinkCache_Users", {PERSISTENT, STRING}},
     {"SunnylinkDongleId", {PERSISTENT, STRING}},
     {"SunnylinkdPid", {PERSISTENT, INT}},
-    {"SunnylinkEnabled", {PERSISTENT, BOOL, "1"}},
+    {"SunnylinkEnabled", {PERSISTENT, BOOL, "0"}},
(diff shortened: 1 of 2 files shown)


sunnyhaibin:
Thanks for the PR.

Given the position we already addressed in #1605, we want to reiterate what we **are** open to: improvements around **onboarding, training, and user education**, for example:
- A clear explanation of what sunnylink provides (remote configuration, device pairing, future features)
- Neutral, factual language that informs users without implying malicious intent
- A mandatory screen during onboarding that allows users to opt out if they choose, with a follow-up confirmation clearly explaining what functionality is lost if sunnylink is disabled

We’re closing this one as well. This proposal still changes the core sunnylink service model (default-off), which we already addressed and declined in #1605. That position hasn’t changed.

To be clear: sunnylink operates on a standard **service-trust / chain-of-trust** model[^1]. Remote services are enabled by default to provide their functionality, and **disabling the service entirely is the explicit opt-out** for users who don’t want that trust model. That option already exists today.

We’re not going to change the default behavior or reframe the service around an adversarial assumption of the operator. Re-submitting variations of the same proposal after it’s been addressed doesn’t move this forward.

If you want to contribute in a way that aligns with the project’s direction, please open a PR focused on onboarding or training only. We’d be happy to review it.

For clarity and consistency with our previously stated position, we won’t be merging this PR as-is.

[^1]: https://community.sunnypilot.ai/t/understanding-the-sunnylink-security-model-overview/1867/2?u=sunnyhaibin

PR #1617, opened 2025-12-26
UI：Change openpilot to sunnypilot and apply the language profile.

Candy0707:


sunnyhaibin (commented):
What difference does this make comparing to the current implementation?

Candy0707:
> 與目前的實現方式相比，這有何不同？

Make events translatable using language
Snipaste_2025-12-27_17-18-01

nayan8teen:
> > 與目前的實現方式相比，這有何不同？
> 
> Make events translatable using language
> Snipaste_2025-12-27_17-18-01
> 

The "sunnypilot" alert was already being translated. Changing that with the replace has no affect imo.

Other events not getting translated should really be an upstream PR instead, as that's a basic upstream functionality that seems to be missing at the moment.

sunnyhaibin:
I'm failing to see the difference it makes with the current changes on `master`. Can you show a before & after screenshots?

PR #1624, opened 2025-12-28
i18n(fr): Add French translations

didlawowo:
Complete French translations for the new Raylib-based UI system. Translates network settings, calibration messages, system dialogs, and various UI elements previously missing from app_fr.po.


didlawowo:
Can someone review it ?

didlawowo:
i have updated the last version with all translation and fixed on screen to have good size of text

PR #1635, opened 2026-01-01
SCC-V: Use p97 for predicted lateral accel

yasu-oh:
**Description**

`SmartCruiseControlVision` previously computed `max_pred_lat_acc` using the maximum of `predicted_lat_accels` (`np.amax`), which can be overly sensitive to single-sample spikes within the prediction horizon. This could lead to transient mispredictions (e.g., brief forward-scene changes) being interpreted as a sharp curve, triggering unintended deceleration constraints.
This PR replaces `np.amax(predicted_lat_accels)` with the 97th percentile (`np.percentile(predicted_lat_accels, 97)`), reducing the impact of one-off outliers while still capturing genuine curvature trends when they persist across multiple samples.

**Verification**

Confirmed the existing `SmartCruiseControlVision` unit tests pass (this change does not affect the current state transition tests).


Discountchubbs:
@yasu-oh I like the idea of using the 97th percentile! Can you add a test with multiple parameters to verify output?

yasu-oh:
Hi, Thanks for the feedback. I’ve added the requested test coverage to validate the 97th-percentile (p97) threshold behavior.

Specifically, the new parameterized tests:

* verify `max_pred_lat_acc` is computed using `np.percentile(..., 97)`,
* cover threshold crossing (p97 > threshold) and non-crossing cases,
* include an outlier “single spike” case to ensure p97 filtering behaves as intended (i.e., max may exceed the threshold while p97 stays below).

All relevant SCC tests are passing locally, including parallel (xdist) runs.

PR #1648, opened 2026-01-16
Sunnylink: block parameter changes when engaged

dzid26:
Blocks any remote modifications while OP is engaged for safety reasons. 
Adds test for those that can result in instant loss of steering. 

Old ref: https://github.com/sunnypilot/sunnypilot/pull/1593#issue-3750214564

### DIFF
--- a/sunnypilot/sunnylink/athena/sunnylinkd.py
+++ b/sunnypilot/sunnylink/athena/sunnylinkd.py
@@ -253,10 +253,17 @@ def getParams(params_keys: list[str], compression: bool = False) -> str | dict[s
 
 @dispatcher.add_method
 def saveParams(params_to_update: dict[str, str], compression: bool = False) -> None:
+  is_engaged = params.get_bool("IsEngaged")
+
   for key, value in params_to_update.items():
     # disallow modifications to blocked parameters
     if key in BLOCKED_PARAMS:
       cloudlog.warning(f"sunnylinkd.saveParams.blocked: Attempted to modify blocked parameter '{key}'")
+      continue
+
+    # Block all params while engaged
+    if is_engaged:
+      cloudlog.warning(f"sunnylinkd.saveParams.blocked_engaged: Attempted to modify '{key}' while engaged")
       continue
 
     try:
--- a/sunnypilot/sunnylink/athena/tests/test_sunnylinkd.py
+++ b/sunnypilot/sunnylink/athena/tests/test_sunnylinkd.py
@@ -4,22 +4,25 @@ Copyright (c) 2021-, Haibin Wen, sunnypilot, and a number of other contributors.
 This file is part of sunnypilot and is licensed under the MIT License.
 See the LICENSE.md file in the root directory for more details.
 """
+import pytest
+
 from openpilot.sunnypilot.sunnylink.athena import sunnylinkd
 
 
 class TestSunnylinkdMethods:
-  def setup_method(self):
+  @pytest.fixture(autouse=True)
+  def setup(self, mocker):
     self.saved_params = []
-
-    self.original_save = sunnylinkd.save_param_from_base64_encoded_string
 
     def mock_save_param(key, value, compression=False):
       self.saved_params.append((key, value, compression))
 
-    sunnylinkd.save_param_from_base64_encoded_string = mock_save_param
+    mocker.patch.object(sunnylinkd, 'save_param_from_base64_encoded_string', mock_save_param)
 
-  def teardown_method(self):
-    sunnylinkd.save_param_from_base64_encoded_string = self.original_save
+    # Mock params with IsEngaged=False by default
+    self.mock_params = mocker.MagicMock()
+    self.mock_params.get_bool.return_value = False
+    mocker.patch.object(sunnylinkd, 'params', self.mock_params)
 
   def test_saveParams_blocked(self):
     blocked_params = {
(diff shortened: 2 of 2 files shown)


sunnyhaibin:
Remote management via sunnylink is already explicitly user-controllable and can be fully disabled via settings.

This PR ignores that existing opt-out and instead introduces a blanket restriction that also impacts legitimate user actions while engaged.

That approach is unnecessary, overbroad, and not aligned with sunnypilot’s and sunnylink’s design. Closing.

dzid26:
I chose broad restriction for simplicity. Would you prefer to narrow it down to param changes that specifically would cause disengagement?

Or you don't recognize the safety issue at all?

devtekve:
> I chose broad restriction for simplicity.

That option *already* exists: simply turn sunnylink **off**. It does not get simpler than that.

> Would you prefer to narrow it down to param changes that specifically would cause disengagement?

This should’ve been **obvious** from the amount of times we’ve rejected PRs[^1][^2][^3] degrading the user experience with sunnylink. I am not sure how you *seriously* thought that preventing users from live-tweaking their params via sunnylink would be anything other than a **degradation** of their experience.

> Or you don't recognize the safety issue at all?

**I do not.** There are already numerous factors that can cause a disengagement; sunnypilot and openpilot mandate **driver attention at all times**, and the driver must be ready to take over immediately. Furthermore, an external actor cannot toggle parameters unless you have *explicitly* paired the device with their account. The **security model handles this[^4]**; there is no safety gap here.

---

If you are **genuinely concerned** about parameters changing while the car is under control, the *correct engineering approach* is to manage that state closer to where the parameter is read, not to block the **transport mechanism**.

You are framing sunnylink as a **bad actor** rather than implementing safeguards against unintended parameter switches at the source. Any service, whether konik stable, konn3kt, or comma connect, could *theoretically* attempt to switch parameters. Attempting to neuter sunnylink specifically is no longer appearing as a safety precaution; it is beginning to look like **malice**. This is becoming a **tiresome waste** of our limited time and resources. Going forward, discuss your approach on our [community forum](https://community.sunnypilot.ai/) **before** opening a PR. This will prevent you from writing code that is *guaranteed* to be rejected.
   
[^1]: https://github.com/sunnypilot/sunnypilot/pull/1593
[^2]: https://github.com/sunnypilot/sunnypilot/pull/1605
[^3]: https://github.com/sunnypilot/sunnypilot/pull/1609
[^4]: **[sunnylink Security Model](https://community.sunnypilot.ai/t/understanding-the-sunnylink-security-model-overview/1867/2)**

dzid26:
Thanks for the taking time to write the honest response. 

I will finish this off and apparently clear up an accusation. 



> Attempting to neuter sunnylink specifically is no longer appearing as a safety precaution; it is beginning to look like malice.

I didn't expect this PR to be controversial. My objective was to improve sunnylink. I think this request is separate to previous ones which mainly were about privacy and consent and they were gracefully rejected for the most part. The safety aspect was never addressed. 

> simply turn sunnylink off. It does not get simpler than that.

Those kind of comments are pointless. In general, they divert from the topic, and they just cause confusion because it's not clear if you want someone to just accept the issue, or you don't agree with the issue. 

> Any service, whether konik stable, konn3kt, or comma connect, could theoretically attempt to switch parameters.

As far as I know, unless they push an update, they cannot remotely disengage. Just like they don't have access over ssh. However, I don't think this is relevant even if you are right.

> I am not sure how you seriously thought that preventing users from live-tweaking their params via sunnylink would be anything other than a degradation of their experience.
>  sunnypilot and openpilot mandate driver attention at all times,

To be snarky as well - I don't know how you can put those two sentences right next to each other. I think tweaking params while engaged is indeed an edge case and principally should be discouraged - unless at standstill. But yeah, of course, I can imagine some other use cases that would justify API use while driving. 


> If you are genuinely concerned about parameters changing while the car is under control, the correct engineering approach is to manage that state closer to where the parameter is read, not to block the transport mechanism.

Pandad service reads shared param state from the filesystem. I am not sure how you would cleanly block it there if this is part of normal interprocess communication.  - I was only interested in blocking remote disengagements. 

> There are already numerous factors that can cause a disengagement; 

I could agree that disengagements should be guarded better on device. I think this is out of scope. 
On a higher level, the disengagement are high probability to be triggered intentionally by the driver per design. Small exception [here](https://github.com/sunnypilot/opendbc/issues/313) where it could be easily triggered accidentally. 

>  the driver must be ready to take over immediately. 

Driver reaction times vary. Assuming optimistic 0.5s reaction and lateral accel of 3m/s2, the vehicle can travel 0.38m laterally due to instant self centering and exceed safety jerk limits. This is enough to cross the center line on some roads and cause a fatal accident. (It doesn't include corrective maneuver to get back on course however). 


> You are framing sunnylink as a bad actor rather than implementing safeguards against unintended parameter switches at the source.

The bad actor could be a third-party getting access to the user browser. The chance is small, but the outcome could be severe. This put burden on the user to keep their security in check.  The alternative is that you just give the user peace of mind.

As far as SL being a bad actor. You guys apply corporate standards to yourself, but as far as the community is concerned you are just strangers on the internet with no expectation to perform background checks, security and access audits. 
In a real world, that we live in, admins can possibly get hacked or just be reckless. Refusing that goes against foss culture and could be a reasonable cause of distrust by some. IMHO.

In contrast, accepting https://github.com/sunnypilot/sunnypilot/pull/1591 was most appreciated, as well as adding the consent and opt-out by default.

PR #1659, opened 2026-01-24
[TIZI/TICI] ui: Fix misaligned turn signals and blindspot indicators with sidebar

Candy0707:
Apply rect center alignment to ensure consistent positioning during state changes.


Candy0707:
image
image

PR #1666, opened 2026-02-01
Fix MADS blocking stock cruise on Chrysler Pacifica Hybrid

rnavarro:
**Description**

When MADS is in lateral-only mode with stock ACC, the cancel logic was incorrectly firing because `CC.enabled=False` while `cruiseState.enabled=True`. This caused stock cruise to disengage
within 70-80ms of engagement.

The fix detects MADS lateral-only mode and skips sending the cancel command, allowing stock ACC to function alongside MADS steering assist.

Scoped to Pacifica Hybrid 2019-25 (tested platform).

**Verification**

- Tested on 2023 Chrysler Pacifica Hybrid
- Before fix: Stock cruise canceled within 70-80ms across multiple attempts
- After fix: Stock cruise stays engaged until user disengages

**Route**

Before fix (bug): https://connect.comma.ai/7d5cf5c92a0349fb/00000040--3f3ae04fe0
After fix (verified): https://connect.comma.ai/7d5cf5c92a0349fb/00000042--6004872eea

Related bug report: https://community.sunnypilot.ai/t/chrysler-pacifica-hybrid-cant-enable-cruise-control/2675

### DIFF
--- a/selfdrive/controls/controlsd.py
+++ b/selfdrive/controls/controlsd.py
@@ -10,6 +10,7 @@ from openpilot.common.realtime import config_realtime_process, DT_CTRL, Priority
 from openpilot.common.swaglog import cloudlog
 
 from opendbc.car.car_helpers import interfaces
+from opendbc.car.chrysler.values import CAR as CHRYSLER_CAR
 from opendbc.car.vehicle_model import VehicleModel
 from openpilot.selfdrive.controls.lib.drive_helpers import clip_curvature
 from openpilot.selfdrive.controls.lib.latcontrol import LatControl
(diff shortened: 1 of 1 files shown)


sunnyhaibin:
Thanks for the PR! Vehicle specific changes must be in https://github.com/sunnypilot/opendbc, not main modules such as controlsd. For controlsd specific modifications/injections, we also have all sunnypilot related changes in controlsd_ext. If any changes must be made, they must be in controlsd_ext, not controlsd.

Feel free to reopen the PRs in the applicable repos and we can review further.

PR #1668, opened 2026-02-02
Fix ICBM/SLA circular dependency for Ford vehicles

alan-polk:

### DIFF
--- a/sunnypilot/selfdrive/controls/lib/speed_limit/speed_limit_assist.py
+++ b/sunnypilot/selfdrive/controls/lib/speed_limit/speed_limit_assist.py
@@ -132,6 +132,18 @@ class SpeedLimitAssist:
       if not self.pcm_op_long and self.is_active:
         return self._speed_limit_final_last
 
+      # ICBM vehicles need target in preActive state to adjust cluster speed
+      # This breaks the circular dependency: SLA preActive -> ICBM adjusts cluster -> SLA becomes active
+      # Non-PCM vehicles get target when active
+      if not self.pcm_op_long and self.is_active:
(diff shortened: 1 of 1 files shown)


sunnyhaibin:
The concern about a circular dependency is incorrect. [`sunnypilot/../longitudinal_planner.py`](https://github.com/sunnypilot/sunnypilot/blob/5c12a7cfc3a199270379d9fb49bd6b309588a844/sunnypilot/selfdrive/controls/lib/longitudinal_planner.py) does not prioritize sources; it simply selects the lowest target using `min(targets)`.

https://github.com/sunnypilot/sunnypilot/blob/5c12a7cfc3a199270379d9fb49bd6b309588a844/sunnypilot/selfdrive/controls/lib/longitudinal_planner.py#L65-L73

When SLA returns `V_CRUISE_UNSET` during SLA's `preActive`, it is effectively filtering itself out as a candidate due to its value (`V_CRUISE_UNSET = 255`). The planner then selects the lowest valid target from the remaining sources (for example, `v_cruise` from the set speed is the lowest out of all speed sources). ICBM consumes this selected target. Since SLA's `preActive` means SLA isn't ready to actuate the new limit yet, allowing planner to follow the current set speed (using the example mentioned, via `v_cruise`) is the correct behavior. Exposing the target early essentially forces actuation before SLA is in `active` state.

With SLA and ICBM decoupled in sunnypilot, and `sunnypilot/opendbc` interface relying on the ICBM messages over cereal to actuate/command can_sends to button press, there shouldn't be any need to change SLA and ICBM controllers on the sunnypilot side.

PR #1695, opened 2026-02-14
[MICI] ui: always offroad

nayan8teen:
image

image

image

image


sunnyhaibin (approved):
LGTM. We could consider having "Always Offroad" status on the home menu in another PR.

zephleggett:
~~I really don't like how this is the first visible item.~~ _Maybe it bugged for me, is it supposed to be last in menu?_
Are we open to putting it in a menu someplace?

PR #1716, opened 2026-02-23
controlsd: fix steer_limited_by_safety not updating under MADS

zephleggett:
I found a MADS-related issue that suppresses the "Turn Exceeds Steering Limit" alert in some cases.
This is a safety issue as warning may not fire when steering saturated under MADS.

## What I noticed

While digging through drive data on my CX-5, I found instances of up to **3.48s of sustained steering saturation at 51 kph** with no "Turn Exceeds Steering Limit" alert firing. The saturation timer never accumulates — `lac.saturated` stays at 0 the entire drive.

## What is happening

When MADS is in lateral-only mode:
- `selfdriveState.active = False`
- `CC.latActive = True`

The `steer_limited_by_safety` update in `publish()` is gated by `selfdriveState.active`:

```python
if self.sm['selfdriveState'].active:    # False under MADS lateral-only
    self.steer_limited_by_safety = ...
```

So once cruise turns off (but MADS lateral stays on), that flag stops updating. If it freezes while torque is being panda rate-limited (mid-ramp), it stays stuck at `True` — which permanently suppresses the saturation timer in `_check_saturation`, since that method skips accumulation when `steer_limited_by_safety` is True.

This looks like it was introduced in #446 (MADS), which correctly updated `CC.latActive` to use `mads.active` via `get_lat_active()`, but missed updating this gate in `publish()`.

## The fix

Use `CC.latActive` instead of `selfdriveState.active` — this already accounts for MADS.

## Testing
I have verified the issue and fix on multiple real drives. I also replayed my existing rlogs to verify all expected alerts are triggering with no false positives.

### DIFF
--- a/common/params_keys.h
+++ b/common/params_keys.h
@@ -190,7 +190,7 @@ inline static std::unordered_map<std::string, ParamKeyAttributes> keys = {
     // Model Manager params
     {"ModelManager_ActiveBundle", {PERSISTENT, JSON}},
     {"ModelManager_ClearCache", {CLEAR_ON_MANAGER_START, BOOL}},
-    {"ModelManager_DownloadIndex", {CLEAR_ON_MANAGER_START | CLEAR_ON_ONROAD_TRANSITION, INT, "0"}},
+    {"ModelManager_DownloadIndex", {CLEAR_ON_MANAGER_START | CLEAR_ON_ONROAD_TRANSITION, INT}},
     {"ModelManager_Favs", {PERSISTENT | BACKUP, STRING}},
     {"ModelManager_LastSyncTime", {CLEAR_ON_MANAGER_START | CLEAR_ON_OFFROAD_TRANSITION, INT, "0"}},
     {"ModelManager_ModelsCache", {PERSISTENT | BACKUP, JSON}},
--- a/selfdrive/car/cruise.py
+++ b/selfdrive/car/cruise.py
@@ -48,14 +48,14 @@ class VCruiseHelper(VCruiseHelperSP):
 
     self.get_minimum_set_speed(is_metric)
 
+    _enabled = self.update_enabled_state(CS, enabled)
+
     if CS.cruiseState.available:
-      _enabled = self.update_enabled_state(CS, enabled)
       if not self.CP.pcmCruise or (not self.CP_SP.pcmCruiseSpeed and _enabled):
         # if stock cruise is completely disabled, then we can use our own set speed logic
         self._update_v_cruise_non_pcm(CS, _enabled, is_metric)
         self.update_speed_limit_assist_v_cruise_non_pcm()
         self.v_cruise_cluster_kph = self.v_cruise_kph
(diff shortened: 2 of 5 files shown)


selfdrive/controls/controlsd.py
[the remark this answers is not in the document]
zephleggett: This is technically correct — when CC.latActive goes False, the flag does retain its last value. But it's not a real problem. When lateral deactivates, state_control() calls self.LaC.reset() which sets sat_time = 0.

I'll implement the suggested change for posterity though.

selfdrive/controls/controlsd.py
```diff
@@ -185,13 +185,15 @@ def publish(self, CC, lac_log):
       hudControl.leftLaneDepart = self.sm['driverAssistance'].leftLaneDeparture
       hudControl.rightLaneDepart = self.sm['driverAssistance'].rightLaneDeparture
 
-    if self.sm['selfdriveState'].active:
+    if CC.latActive:
```
sunnyhaibin: Switching from `selfdriveState.active` directly to `carControl.latActive` may cause unintended behaviors, as now we're also including checks such as steering faults or standstill allowed conditions here.

Instead, I would use the existing method that we have in `controlsd_ext` in here: https://github.com/sunnypilot/sunnypilot/blob/111e8897be79f4c99734e77d0c389612f14a8e50/sunnypilot/selfdrive/controls/controlsd_ext.py#L57-L66

This is also used in `controlsd`:
https://github.com/sunnypilot/sunnypilot/blob/111e8897be79f4c99734e77d0c389612f14a8e50/selfdrive/controls/controlsd.py#L111-L112

selfdrive/controls/controlsd.py
```diff
@@ -185,13 +185,15 @@ def publish(self, CC, lac_log):
...
       CO = self.sm['carOutput']
       if self.CP.steerControlType == car.CarParams.SteerControlType.angle:
         self.steer_limited_by_safety = abs(CC.actuators.steeringAngleDeg - CO.actuatorsOutput.steeringAngleDeg) > \
                                               STEER_ANGLE_SATURATION_THRESHOLD
       else:
         self.steer_limited_by_safety = abs(CC.actuators.torque - CO.actuatorsOutput.torque) > 1e-2
+    else:
+      self.steer_limited_by_safety = False
```
sunnyhaibin: Not necessarily needed here as this should only run when lateral control is **active**.
```suggestion
```

selfdrive/controls/controlsd.py
zephleggett: ty for the suggestion! I am using lat_active now/

selfdrive/controls/controlsd.py
zephleggett: Ah, this was suggested by CoPilot PR Review.
I will remove.

sunnyhaibin (approved):
Great fix. Thanks for the PR!

PR #1720, opened 2026-02-25
pandad: remove unsupported panda checks

sunnyhaibin:
This was a dependency for xnor's Rivian Longitudinal Upgrade. No longer needed after https://github.com/sunnypilot/sunnypilot/pull/1712


sunnyhaibin:
Still needed for now.

PR #1726, opened 2026-02-27
opendbc: update submodule for KIA NQ5 Sportage fuzzy fingerprint support

syrull:
## Description

Updates the `opendbc_repo` submodule to include sunnypilot/opendbc#427, which adds `KIA_SPORTAGE_5TH_GEN` (NQ5) to `CANFD_FUZZY_WHITELIST`.

## Problem

The KIA Sportage 5th Gen (NQ5) is a CAN-FD car that is blacklisted from fuzzy fingerprinting. Cars with firmware versions not in the exact fingerprint database show as "unsupported car" on the device, even though the NQ5 platform is otherwise fully supported.

## Dependency

- Requires sunnypilot/opendbc#427 to be merged first, then this submodule pointer should be updated to point to the merged commit on `sunnypilot/opendbc`.


sunnyhaibin:
This PR changes the opendbc submodule to the repo that is not maintained by sunnypilot. All opendbc changes need to be merged on sunnypilot/opendbc, then the changes will be bumped when applicable.

PR #1740, opened 2026-02-28
ui/mici: [last] wire sunnypilot settings panels into top-level navigation

zephleggett:
## Summary
- Wire all SP settings panels into the top-level settings navigation
- Uses upstream `SettingsBigButton` with `gui_app.texture()` for icons
- Inserts SP panels (sunnylink, models, cruise, steering, display, visuals, trips) after the first upstream button

## Test plan
- [ ] Run `PYTHONPATH=. BIG=0 .venv/bin/python selfdrive/ui/ui.py`
- [ ] Verify all 7 SP panels appear in settings and navigate correctly
- [ ] Verify upstream panels (toggles, network, device, developer) still work


zephleggett:
Updated:
- Rebased onto current master (removed stale submodule diff)
- All user-visible strings now wrapped with `tr()` for localization
- Single clean commit, only the intended layout file changed

Note: sentry bot import errors are expected — this PR depends on #1734 (shared-widgets) being merged first.

PR #1752, opened 2026-03-03
Rivian: Flash xnor's Longitudinal Upgrade Kit prior supported panda check

lukasloetkolben:
Flash external Black Panda for Rivian longitudinal control before the stock panda setup runs.
This filters out the Black Panda from the serial list so the normal flash and health check logic only sees the internal panda. Fixes a loop where the Black Panda was picked up first, causing "internal panda is missing" errors. This was due to the line `panda_serial = panda_serials[0]`


sunnyhaibin:
https://community.sunnypilot.ai/t/longitudinal-support-issues/3449/3?u=sunnyhaibin

PR #1767, opened 2026-03-15
[MICI] ui: detailed onroad view

rav4kumar:
image

https://github.com/user-attachments/assets/c571fac5-e897-46e5-b6e3-31042883aefc


Discountchubbs:
Post SLA on left, master on right. 95 degree ambient temperature for right, however this review shows stability in both branches and thus, this PR passes on memory, cpu, and execution usage.

Screenshot 2026-03-17 at 8 16 11 PM
Screenshot 2026-03-17 at 8 15 08 PM
Screenshot 2026-03-17 at 8 14 06 PM

PR #1779, opened 2026-03-25
NNLC: restore pre-v1 PID gains in torque extension

mmmorks:
**Description**

The v1 torque controller refactor changed the NNLC extension's PID gains from the original per-vehicle defaults (kp=1.0, ki=0.3, kf=1.0) to the base controller's new values (kp=0.8, ki=0.15, speed-interpolated, no kf).

The NNLC extension operates in torque space with its own independent PID loop, so its gains should not have been coupled to the base controller's lateral acceleration PID. This results in weaker steering response and ping-pong oscillation for NNLC users, with no workaround since Enforce Torque Lateral Control and NNLC are mutually exclusive.

This restores the PID gains to match the behavior from v2025.003.000 and earlier.

Related: https://community.sunnypilot.ai/t/torque-lateral-control-update-tuning-toggle-v0-vs-v1/3124

**Verification**

Tested on Genesis GV70 Electrified with NNLC enabled. Confirmed reduced steering oscillation and restored steering feel matching v2025.003 behavior. No change when NNLC is disabled (base controller unaffected).


AntoineGS:
Just showing my interest in this getting merged as on the Ioniq 5 things are pretty bad with the ping pong-ing!

sunnyhaibin (approved):

sunnyhaibin:
@mmmorks  Nice fix! Thanks for the PR!

PR #1782, opened 2026-03-27
[MICI] ui: align upstream changes with sunnypilot settings buttons

nayan8teen:


zephleggett:
**Correction on my previous comment** — I was comparing against a downstream branch, not current master. On current `master`:

- `_scroller._items` is still the correct attribute (no public `.items` property yet) ✓
- `SunnylinkLayoutMici` still accepts `back_callback` ✓

So the PR's changes look correct against master. Apologies for the noise.

sunnyhaibin (approved):

PR #1798, opened 2026-04-12
[MICI] ui: add sunnylink info & connectivity check

nayan8teen:


devtekve (dismissed):
Looks good on device, the sponsor tier for me was “free” so we may need to fix the recognition there but that’s not introduced on this PR.

I didn’t test yet the offline because I have a sim so it doesn’t make it easy lol

devtekve (approved):
working as expected <3

devtekve (approved):

PR #1811, opened 2026-04-20
ui/networking: implement tethering NAT as WifiManagerSP

cloudrainstar:
**Description**

Reimplements the tethering NAT fix from https://github.com/sunnypilot/sunnypilot/pull/1750 as a sunnypilot hook.

This adds `WifiManagerSP` and uses it in the sunnypilot settings path instead of modifying upstream `WifiManager` directly. `WifiManagerSP` adds the `iptables-legacy` MASQUERADE rule when tethering is enabled with `_ipv4_forward`, and removes it when tethering is disabled.

This new fix also addresses Sentry bot's concerns in https://github.com/sunnypilot/sunnypilot/pull/1750 regarding the rule persisting after deactivating tethering.

~~This branch is built from a clean master with only the two relevant commits, but requires the ToggleSP fix https://github.com/sunnypilot/sunnypilot/pull/1750/changes/d34155ccb973365184ff1173d2f6f8b9c4ebce98 from https://github.com/sunnypilot/sunnypilot/pull/1750 to work properly.~~ Edit: This should work since https://github.com/sunnypilot/sunnypilot/pull/1810 is merged.

Please also note that this is scoped to the sunnypilot settings path only. The mici path still uses upstream `WifiManager` directly and I don't have a comma 4 to test this.

**Verification**

Tested on Comma 3X:
- enable tethering from UI
- confirm NAT rule is added
- disable tethering
- confirm NAT rule is removed
- repeated toggles do not create duplicate rules


sunnyhaibin (commented):
1. Did you consider cleaning up stale iptables rules on startup?                                          
2. Was dropping `time.sleep(5)` intentional? Upstream used it to race against NM auto-enabling `ip_forward` on AP activation. Without it the `sysctl=0` call might fire before NM touches the flag.

cloudrainstar:
> 1. Did you consider cleaning up stale iptables rules on startup?
> 2. Was dropping `time.sleep(5)` intentional? Upstream used it to race against NM auto-enabling `ip_forward` on AP activation. Without it the `sysctl=0` call might fire before NM touches the flag.

1. I assumed cold boot starts clean because we do not persist iptables state. However, I guess it would not necessarily be clean across process/UI restarts in the same OS session. A defensive cleanup on init should work. Let me add it.
2. Dropping time.sleep(5) was not intentional. Upstream’s rationale still applies for the _ipv4_forward == false path, so it definitely makes sense to do that. I’ll restore it in another commit.

cloudrainstar:
Tested on C3X, tethering works as expected on this branch. Ready for your review again.

devtekve:
@cloudrainstar could I bother you asking for a toggle for this? as in "Allow hotspot to share internet" or something like that? Mainly so that if a user has issues of any kind, they can just toggle it off and should be back to close as stock

cloudrainstar:
> @cloudrainstar could I bother you asking for a toggle for this? as in "Allow hotspot to share internet" or something like that? Mainly so that if a user has issues of any kind, they can just toggle it off and should be back to close as stock

Reworked this to remove the previous hook from `Enable Tethering` entirely.

The new behavior is an explicit `Use Legacy NAT` toggle under tethering settings, instead of automatically
applying the `iptables-legacy` fallback when tethering is enabled. That keeps stock behavior unchanged unless the
user opts into the workaround.

Current logic:
- `Use Legacy NAT` is only visible when tethering is active and the upstream path would allow forwarding
(`PrimeType.NONE` / `PrimeType.LITE`)
- when tethering is not active, the toggle is hidden
- when tethering is enabled again later, `Use Legacy NAT` starts off again by default

That default-off behavior is intentional to keep the change minimal relative to stock. Enabling tethering alone
does not introduce any additional NAT workaround. The user has to explicitly toggle `Use Legacy NAT` on before
behavior changes.

The toggle state tracks live rule state rather than just local UI state, and there is a periodic re-check while
eligible so the UI can reconcile with actual `iptables-legacy` state if it drifts.

Tested on Comma 3X:
1. `Enable Tethering` on, `Use Legacy NAT` off:
   - hotspot activates
   - tethered devices do not get internet access
2. `Use Legacy NAT` on:
   - tethered devices can access the internet
3. `Use Legacy NAT` off again:
   - tethered devices lose internet access
4. `Use Legacy NAT` on, then `Enable Tethering` off:
   - tethering deactivates
   - `Use Legacy NAT` hides itself
5. `Enable Tethering` on again:
   - `Use Legacy NAT` is shown again in the off state

So the fallback now requires explicit user action each session, which is in line with keeping behavioral changes
minimal.

cloudrainstar:
Changed TETHERING_UPLINK_IFACE from "wwan0" to "ppp0" due to commit 1a93104b from upstream.

PR #1813, opened 2026-04-24
[tizi/tici] ui: Camera offset controls

CHaucke89:
Just a little something for comma 3/3X users. Adds controls for the camera offset to the Models settings panel using the same logic as sunnylink, so it will not be shown unless the model is changed from the default.

disabled
enabled
enabled_desc


sunnyhaibin (approved):
LGTM. Thanks for giving TIZI/TICI some love ❤️

PR #1814, opened 2026-04-25
sunnylink: Remove unused API endpoint

devtekve:
This pull request removes the `ws_queue` functionality from the Sunnylink Athena service, simplifying the threading model and related API code. The main impact is the elimination of the resume queue logic, which previously attempted to resume queued tasks with retries.

**Removal of resume queue logic:**

* The `ws_queue` thread and its associated function have been removed from the Athena service, so the system no longer attempts to resume queued operations in a background thread. (`sunnypilot/sunnylink/athena/sunnylinkd.py`) [[1]](diffhunk://#diff-40ad7bc50889c2d969b56606b55c2eccf4de1253185eba99a149ba10c70d01beL67) [[2]](diffhunk://#diff-40ad7bc50889c2d969b56606b55c2eccf4de1253185eba99a149ba10c70d01beL150-L180)
* The `resume_queued` method has been deleted from the `SunnylinkApi` class, as it is no longer needed. (`sunnypilot/sunnylink/api.py`)

### DIFF
--- a/sunnypilot/sunnylink/api.py
+++ b/sunnypilot/sunnylink/api.py
@@ -30,10 +30,6 @@ class SunnylinkApi(BaseApi):
 
     return super().api_get(endpoint, method, timeout, access_token, session, json, **kwargs)
 
-  def resume_queued(self, timeout=10, **kwargs):
-    sunnylinkId, commaId = self._resolve_dongle_ids()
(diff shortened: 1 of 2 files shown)


devtekve:
@copilot do a quick double check to make sure we arent leaving anything related to resume queued behind on sunnylinkd

sunnyhaibin (approved):

PR #1815, opened 2026-04-25
Cam and log sync with rsync and google drive

techfoundrynz:


techfoundrynz:
Oops, PR into the wrong repo

PR #1816, opened 2026-04-25
SCC-M: fix operator precedence in quadratic roots

0x3k:
**Description**

Fixes a bug in the Smart Cruise Control map controller's quadratic root calculation.

`map_controller.py:154-155` computed the time to reach a moderate-curve target velocity as:
```python
t_a = -1 * ((b**2 - 4 * a * c) ** 0.5 + b) / 2 * a
t_b = ((b**2 - 4 * a * c) ** 0.5 - b) / 2 * a
```

In Python `/` and `*` have equal precedence and are left-associative, so `... / 2 * a` parses as `(... / 2) * a` (the result is multiplied by `a` instead of divided by `2a`), so the denominator needs to be parenthesized.

With a = 0.5 * TARGET_JERK (small), the bug scaled t by roughly a^2, making it about two orders of magnitude too small. As a result `max_d` was tiny and the moderate-curve branch (tv > min_accel_v) effectively never tripped so the controller wouldn't slow for upcoming curves until they were already much closer than intended.

**Verification**

- Added regression test test_moderate_curve_quadratic_root that exercises the moderate-curve branch with a 40 m waypoint at v_ego=25, a_ego=0, tv=24. With the fix max_d ≈ 45 m (waypoint detected, v_target == 24); with the old code max_d ≈ 4 m (waypoint missed, assertion fails).
- Existing test_map_controller.py cases still pass.


zephleggett:
This looks correct to me, nice find. I just got ICBM working properly on mazdas and will test.

sunnyhaibin (approved):
LGTM. Thanks for the improvements!

PR #1819, opened 2026-04-26
Sync: `commaai/openpilot:master` → `sunnypilot/sunnypilot:master`

sunnyhaibin:

PR #1837, opened 2026-05-10
ui: show default model name

nayan8teen:
image

image

image

sunnylink pr: http://github.com/sunnypilot/sunnylink-frontend/pull/79

### DIFF
--- a/common/model.h
+++ b/common/model.h
@@ -1,1 +1,0 @@
-#define DEFAULT_MODEL "POP model (Default)"
--- a/common/model.py
+++ b/common/model.py
@@ -1,1 +1,0 @@
-DEFAULT_MODEL = "CD210 Model (Default)"
--- a/common/model.py
+++ b/common/model.py
@@ -1,0 +1,1 @@
+DEFAULT_MODEL = "CD210 Model (Default)"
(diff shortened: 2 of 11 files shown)


sunnyhaibin:
Instead of relying on the param for the default model name, we could add a new field to the settings schema, then have the frontend read that.

`master` currently also runs POP model as well :D

sunnyhaibin (approved):

nayan8teen:
> Instead of relying on the param for the default model name, we could add a new field to the settings schema, then have the frontend read that.
> 
> `master` currently also runs POP model as well :D

Didn't we sync after the POP revert? 🤦🏻

sunnyhaibin:
> > Instead of relying on the param for the default model name, we could add a new field to the settings schema, then have the frontend read that.
> > `master` currently also runs POP model as well :D
> 
> Didn't we sync after the POP revert? 🤦🏻

We haven’t merged it yet 😆 https://github.com/sunnypilot/sunnypilot/pull/1819

nayan8teen:
> > > Instead of relying on the param for the default model name, we could add a new field to the settings schema, then have the frontend read that.
> > > `master` currently also runs POP model as well :D
> > 
> > Didn't we sync after the POP revert? 🤦🏻
> 
> We haven’t merged it yet 😆 https://github.com/sunnypilot/sunnypilot/pull/1819

Well. Who's fault is that 🤣

PR #1842, opened 2026-05-17
sunnylink SDUI: tweak DisableUpdate param for clarity

sunnyhaibin:


devtekve:
@sunnyhaibin does the tici UI also convey the same?

PR #1855, opened 2026-06-06
modeld_v2: safe model validation

Discountchubbs:
Better model validation. This checks the hash and physical discrepancy's of the model loaded on device. If it doesn't match the current selector version, it will be forced to stock modeld. If the model somehow doesn't exist locally, it will be be forced to stock modeld. If the model is corrupt, it will go to default. If the model is somehow deleted it'll go to default model. I hypothesize that the refactor made in helpers.py will allow a more controlled experience for the userbase.

Screenshot 2026-06-06 at 9 03 05 AM

Screenshot 2026-06-06 at 9 05 09 AM

https://github.com/user-attachments/assets/0572b1a5-417b-43c2-9a1c-a2109d22fc5b


sunnyhaibin (approved):
Big red diff!

PR #1860, opened 2026-06-08
Revert "deprecate `carState.brake`" for Honda Gas Interceptor

mvl-boston:
Required to support https://github.com/sunnypilot/opendbc/pull/481


sunnyhaibin (approved):
Thanks!

PR #1867, opened 2026-06-24
fix: Sunnylink settings panel stale display and offroad write enforcement

Geomglot:
## Purpose

Two related bugs in the Sunnylink remote settings path:

1. **Settings panel shows stale value after a remote change** — multiple-button and toggle widgets did not re-read their param when a settings panel was re-opened. If Sunnylink wrote a new value remotely, the panel continued to show the old value until the UI process restarted.

2. **Sunnylink could write offroad-only params while onroad** — settings marked `offroad_only` in the schema (greyed out on-device while driving) had no enforcement on the remote write path. `sunnylinkd.saveParams()` accepted those writes regardless of drive state.

Both fixes affect all supported vehicles.

Originally proposed/discussed: https://community.sunnypilot.ai/t/changes-via-sunnylink-require-device-reboot-to-be-effective/5400

## Changes

- `system/ui/sunnypilot/widgets/list_view.py` — `MultipleButtonActionSP.show_event()` re-reads its param on panel open
- `system/ui/sunnypilot/widgets/toggle.py` — `ToggleSP.show_event()` re-reads its param on panel open
- `sunnypilot/sunnylink/athena/sunnylinkd.py` — `saveParams()` loads `offroad_only` params from `settings_ui.json` at startup and silently drops writes to those params while `IsOnroad` is true; `ParamsVersion` is only incremented when at least one write succeeds
- `sunnypilot/sunnylink/docs/README.md` — documents the new enforcement behaviour

## Verification

- Changed a multiple-button setting via Sunnylink web app while device was parked; reopened the settings panel on-device — new value displayed immediately without restarting the UI.
- Attempted to push an offroad-only setting via Sunnylink while driving — param value did not change; rejection logged by `sunnylinkd`.
- Pushed the same offroad-only setting while parked — write succeeded normally.
- Normal on-device toggle and multiple-button interaction unaffected.

### DIFF
--- a/sunnypilot/sunnylink/athena/sunnylinkd.py
+++ b/sunnypilot/sunnylink/athena/sunnylinkd.py
@@ -57,6 +57,58 @@ BLOCKED_PARAMS = {
   "ParamsVersion",         # Device-managed version counter
 }
 
+_SETTINGS_UI_PATH = os.path.join(os.path.dirname(os.path.dirname(os.path.abspath(__file__))), "settings_ui.json")
+
+
+def _extract_rule_params(schema: dict, rule_type: str) -> frozenset[str]:
+  result: set[str] = set()
+
+  def has_rule(rules: list) -> bool:
+    for r in rules:
+      if not isinstance(r, dict):
+        continue
+      if r.get('type') == rule_type:
+        return True
+      for v in r.values():
+        if isinstance(v, list) and has_rule(v):
+          return True
+    return False
+
+  def scan(items: list) -> None:
+    for item in items:
+      if not isinstance(item, dict):
+        continue
+      key = item.get('key')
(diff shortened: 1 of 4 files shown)


Geomglot:
The 5 failing checks (`build release`, `build macOS`, `Create UI Report`, `unit tests`, `preview`) all fail at the `scons` build step with `Unexpected non-vendored library 'x264'` / `'avformat'` before any of our code is reached. The same error appears in #1811 (unrelated PR, same timeframe) and in today's scheduled `prebuilt` runs on master — this appears to be a CI infrastructure issue on the sunnypilot side rather than a problem with these changes.

`static analysis` passed. Happy to re-run the checks once the runner environment is fixed.

PR #1873, opened 2026-07-07
2015 Honda Accord v6L AU Support

SoRadGaming:
This PR adds Support for 2013 - 2015 Honda Accords in Australia that came with ACC and LKAS [TECH & V6L Trims]

MVL and I have already got OP Longitudinal Controls working, moved over to sunny-pilot now that the port is more viable, and need comma pedal and serial steering board support added.

**Checklist**

- [x] added entry to CAR in selfdrive/car/*/values.py and ran `selfdrive/car/docs.py` to generate new docs
- [x] test route added to [routes.py](https://github.com/commaai/openpilot/blob/master/selfdrive/car/tests/routes.py)
- [x] route with openpilot:
- [x] route with stock system:
- [ ] car harness used (if comma doesn't sell it, put N/A):
- [x] Fix TSA Error
- [x] Map Radar Bits
- [ ] Comma Pedal Support
- [x] Re-calculate Nidec Values for Elesys Car (Brakes)
- [ ] Add Support for Serial Steering
- [ ] Test Stock AEB on MAIN off

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -3,7 +3,8 @@
   url = https://github.com/sunnyhaibin/panda.git
 [submodule "opendbc"]
   path = opendbc_repo
-  url = https://github.com/sunnypilot/opendbc.git
+  url = https://github.com/SoRadGaming/opendbc.git
+  branch = sp-master
 [submodule "msgq"]
   path = msgq_repo
   url = https://github.com/commaai/msgq.git


SoRadGaming:
I'm assuming need to open the PR on opendbc repo, still working on it for now, will fix the PR once its ready for reveiw

PR #1885, opened 2026-07-25
speed_limit: fix GPS fix-age clock-domain bug (unreachable map adaptation)

spawahh:
**Description**

`SpeedLimitResolver` computed GPS fix age by subtracting the GPS message's wall-clock `unixTimestampMillis` (epoch, ~1.7e12 ms) from `time.monotonic()` (uptime seconds):

```python
gps_fix_age = time.monotonic() - gps_data.unixTimestampMillis * 1e-3
```

The result is always hugely negative, which breaks two things:

1. **The staleness guard can never fire**, so stale map data is always accepted:

   ```python
   if gps_fix_age > LIMIT_MAX_MAP_DATA_AGE:   # never True
       return
   ```

2. **`distance_since_fix` is always negative**, so `distance_to_speed_limit_ahead` stays inflated and the upcoming-lower-speed-limit adaptation branch is unreachable — which is the cause behind the existing `FIXME-SP: this is not working as expected` sitting on that very branch.

Measured on a build (`LIMIT_MAX_MAP_DATA_AGE = 10.0`, `v_ego = 15 m/s`, ahead = 200 m):

```
buggy gps_fix_age        = -1,784,937,636 s     (~ -56 years)
staleness guard fires?   = False                 <- stale data never rejected
distance_since_fix       = -26,774,064,536 m     (physically impossible)
distance_to_limit_ahead  = 26,774,064,736 m      (inflated from 200)
```

This looks like a blind `time.time` -> `time.monotonic` swap to satisfy the ruff ban on `time.time`, which silently changed clock domains.

**Fix:** measure GPS age in the monotonic domain via `sm.recv_time[gps_service]` (consumer-side receipt time), which also stays large until the first fix arrives, so a cold start is still correctly treated as stale. Both call sites now share one `_gps_fix_age()` helper, and the stale `FIXME-SP` comment is removed.

**Verification**

Unit tests, on Ubuntu 24.04 / Python 3.12.3, against a tree containing current master plus only this commit:

```
test_speed_limit_resolver.py             : 24 passed before -> 25 passed after
whole speed_limit/ directory             : 45 passed  (includes test_speed_limit_assist.py)
sunnypilot/selfdrive/controls/lib/tests  : 56 passed
```

This adds `test_upcoming_lower_limit_triggers_adaptation` covering the previously unreachable branch, and updates existing tests that had masked the bug by feeding monotonic-valued timestamps into `unixTimestampMillis`.

I also decoded a 24-minute drive from my own car (comma 3X; `liveMapDataSP` at ~1 Hz, 1411 messages, 1389 with `speedLimitValid`) and recomputed the unpatched expression against the logged GPS messages:

```
buggy gps_fix_age      = -1,784,859,278 s   (constant across the drive)
staleness guard fired  = 0 / 1399 GPS messages (0.0%)
```

So on a real drive the guard never rejected map data once — it cannot, by construction. The dead adaptation branch also had real work available:

```
frames with speedLimitAheadDistance > 0                       : 310
of those, upcoming limit < current speed (adaptation-eligible): 150
```

At ~1 Hz that is roughly 150 seconds of that drive where an upcoming lower limit was known and the adaptation branch should have been engaging but could not. Map limits changed only 7 times across the drive, each value held for minutes, so a stale value has a long window in which to persist undetected.

To be straight about the limits of my verification:

- The numeric demonstration and the unit tests above are the evidence.
- I have **not** road-tested the adaptation behaviour in Speed Limit Assist mode. My car is a Subaru, which needs alpha longitudinal for Assist, and I don't run that. The resolver itself runs in `plannerd` on every car regardless of mode, so the bug isn't platform-specific — but someone on a factory-longitudinal car should confirm the on-road behaviour before this is relied on.
- Note this makes a previously dead code path live: the car will now begin slowing for an upcoming lower limit rather than at it. That's the intended behaviour, but it is a behaviour change on a live driving feature and deserves a deliberate test drive.

**Not included here**

`LongitudinalPlannerSP.__init__` assigns `self.resolver = SpeedLimitResolver()` twice (lines 27 and 30) — harmless, but one is dead. I left it out to keep this PR single-purpose; happy to open a separate one if you'd like it cleaned up.

### DIFF
--- a/openpilot/sunnypilot/selfdrive/controls/lib/speed_limit/speed_limit_resolver.py
+++ b/openpilot/sunnypilot/selfdrive/controls/lib/speed_limit/speed_limit_resolver.py
@@ -119,12 +119,23 @@ class SpeedLimitResolver:
     self._reset_limit_sources(SpeedLimitSource.map)
     self._process_map_data(sm)
 
+  def _gps_fix_age(self, sm: messaging.SubMaster) -> float:
+    # Time since the last GPS message, in the monotonic domain. The message's own
+    # unixTimestampMillis is wall-clock epoch (ms) and must not be mixed with
+    # time.monotonic(); recv_time is the consumer-side monotonic receipt time and
+    # stays large (data treated as stale) until the first fix arrives.
+    recv_time = getattr(sm, 'recv_time', None)
+    if recv_time is None:
+      # The longitudinal maneuver harness (selfdrive/test/longitudinal_maneuvers/plant.py)
+      # passes a plain dict instead of a SubMaster, so there is no receipt time to read.
+      # It synthesizes a fresh message every step, so treat the data as current.
+      return 0.
+    return time.monotonic() - recv_time[self._gps_location_service]
+
   def _process_map_data(self, sm: messaging.SubMaster) -> None:
-    gps_data = sm[self._gps_location_service]
     map_data = sm['liveMapDataSP']
 
-    gps_fix_age = time.monotonic() - gps_data.unixTimestampMillis * 1e-3
-    if gps_fix_age > LIMIT_MAX_MAP_DATA_AGE:
+    if self._gps_fix_age(sm) > LIMIT_MAX_MAP_DATA_AGE:
       return
 
     speed_limit = map_data.speedLimit if map_data.speedLimitValid else 0.
@@ -133,16 +144,14 @@ class SpeedLimitResolver:
     self._calculate_map_data_limits(sm, speed_limit, next_speed_limit)
 
   def _calculate_map_data_limits(self, sm: messaging.SubMaster, speed_limit: float, next_speed_limit: float) -> None:
-    gps_data = sm[self._gps_location_service]
     map_data = sm['liveMapDataSP']
 
-    distance_since_fix = self.v_ego * (time.monotonic() - gps_data.unixTimestampMillis * 1e-3)
+    distance_since_fix = self.v_ego * self._gps_fix_age(sm)
     distance_to_speed_limit_ahead = max(0., map_data.speedLimitAheadDistance - distance_since_fix)
 
     self.limit_solutions[SpeedLimitSource.map] = speed_limit
     self.distance_solutions[SpeedLimitSource.map] = 0.
 
-    # FIXME-SP: this is not working as expected
     if 0. < next_speed_limit < self.v_ego:
       adapt_time = (next_speed_limit - self.v_ego) / LIMIT_ADAPT_ACC
       adapt_distance = self.v_ego * adapt_time + 0.5 * LIMIT_ADAPT_ACC * adapt_time ** 2
--- a/openpilot/sunnypilot/selfdrive/controls/lib/speed_limit/tests/test_speed_limit_resolver.py
+++ b/openpilot/sunnypilot/selfdrive/controls/lib/speed_limit/tests/test_speed_limit_resolver.py
@@ -55,6 +55,10 @@ def setup_sm_mock(mocker: MockerFixture):
     'carStateSP': car_state_sp,
     'gpsLocation': gps_data,
   }[key]
+  # GPS freshness is measured from the monotonic receipt time; a recent value keeps
+  # the map data considered fresh.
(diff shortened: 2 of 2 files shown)


spawahh:
Pushed a fix for the CI failure — thanks for the fast feedback from the run.

**What broke:** all 92 failures were one root cause, `AttributeError: 'dict' object has no attribute 'recv_time'`, across `test_cruise_speed.py` (12), `test_following_distance.py` (18) and `test_longitudinal.py` (58 subtests). They all go through `selfdrive/test/longitudinal_maneuvers/plant.py`, which passes a **plain dict** as `sm` rather than a `SubMaster`:

```python
sm = {'radarState': radar.radarState, ..., 'liveMapDataSP': ..., 'gpsLocation': ...}
```

So item access works but attribute access doesn't. The original buggy line happened to be dict-compatible (`sm['gpsLocation'].unixTimestampMillis`), which is why this never surfaced before; reading `sm.recv_time` is not. My fault for not running those suites.

**Fix:** `_gps_fix_age` now tolerates a non-SubMaster `sm`:

```python
recv_time = getattr(sm, 'recv_time', None)
if recv_time is None:
  # plant.py passes a plain dict instead of a SubMaster, so there is no receipt
  # time to read. It synthesizes a fresh message every step, so treat it as current.
  return 0.
return time.monotonic() - recv_time[self._gps_location_service]
```

Onroad `sm` is always a real `SubMaster`, so the fallback is harness-only and the staleness guard is unchanged in production. Verified locally that a `SubMaster`-like object with a stale `recv_time` (999 s) is still rejected, and that a plant-style dict no longer raises.

Still green locally: whole `speed_limit/` directory 45 passed, `sunnypilot/selfdrive/controls/lib/tests` 56 passed.

One caveat in the interest of being straight: I could not run the three failing suites locally — they need the acados MPC codegen and `scons` won't complete in my environment. The fix is verified at the exact failure point rather than by executing those tests, so I'm relying on this CI run to confirm all 92.

spawahh:
**Follow-up: the adaptation branch is now confirmed firing on a real road.**

My earlier evidence was a numeric demonstration plus log analysis showing the staleness guard could never fire. I can now show the revived adaptation branch actually engaging on public road, on a comma 3X running this patch.

Route `00000060--a8f773ce3e` (11 segments):

```
distToSpeedLimit > 0 : 100 frames, max 49.3 m
examples (dist_m, upcoming_limit_mph, v_ego_mph):
  (46.3, 20, 29.0)  (45.7, 20, 29.1)  (45.0, 20, 29.2)  (44.4, 20, 29.2)
```

`distToSpeedLimit` is only assigned inside the adaptation branch, so a non-zero value is proof the branch executed — which was impossible before this fix, since `distance_to_speed_limit_ahead` was inflated by a hugely negative `distance_since_fix`.

The magnitudes match the kinematics. With `LIMIT_ADAPT_ACC = -1.0 m/s²`, adapting from 29 mph down to a 20 mph limit requires 44.1 m (≈56 m at 31 mph), and firing began at 46–49 m.

Two further drives confirm it also stays quiet when it should:

| Route | Branch-eligible frames | Distance to limit change | Fired |
|---|---|---|---|
| `00000060--a8f773ce3e` | 40 | 43–49 m | yes (100 frames) |
| `00000061--78c660b2aa` | 15 | 774–815 m | no |
| `00000062--ff7cf4f8de` | 44 | 636–691 m | no |

Both non-firing cases had a valid upcoming lower limit but were far beyond the adaptation distance, so declining to adapt is the correct outcome rather than a miss.

All three drives ran with Speed Limit Assist **disabled** (`assist.state = disabled` in every logged frame), so the resolver output was informational only and did not affect vehicle control. That remains the limit of what I can demonstrate: the branch computes correct adaptation distances on road, but I have not exercised the resulting speed control in Assist mode — my car needs alpha longitudinal for that and I don't run it.

sunnyhaibin:
The routes provided have no dongle ID so it's not even verifiable. Whether the fix is valid or not, the whole PR is obviously LLM generated and we can't review that. Resubmit in your own words with verifiable evidence if the bug is real.

PR #1906, opened 2026-08-12
LagdToggle: fix inverted get_lat_delay branches

RWL-Dittrich:
**Description**

`get_lat_delay` had its two branches the wrong way round.

All callers pass the value lagd publishes as `stock_lat_delay`, while `LagdToggle.update()` only writes the fixed `steerActuatorDelay + software delay` sum into `LagdValueCache` when the toggle is off. So with live learning off, `get_lat_delay` returned the published learner value and ignored the cache, which is where the configured delay actually lives.

Two effects:

- "Adjust Software Delay" had no effect at all.
- Turning off "Live Learning Steer Delay" still used the learner's output, including its unconverged `steerActuatorDelay + 0.2` seed from `lagd.py`.

It's masked at default settings, since `LagdToggleDelay` defaults to 0.2 and lagd's seed offset is also 0.2, so both branches return the same number until the slider is moved.

**Verification**

Called `get_lat_delay` directly on the device with a stub `Params` for both toggle states, using `stock_lat_delay=0.55` and `LagdValueCache=0.40`.

Before:

```
LagdToggle=True  -> 0.400  (cache)
LagdToggle=False -> 0.550  (published)
```

After:

```
LagdToggle=True  -> 0.550  (published)
LagdToggle=False -> 0.400  (cache)
```
Note: this fix does change behaviour for anyone running with live learning off and a software delay other than 0.2, since they'll start getting the delay they configured.


nayan8teen (approved):
Looks like this got inverted during a recent refactor and went unnoticed. 

Thanks for the PR!!

PR #1918, opened 2026-08-18
modeld_v2: fix features_buffer alignment for supercombo models

### LINKED ISSUE #1917: modeld_v2: new supercombo models have misaligned features_buffer
### Describe the bug

When upstream combined the driving model into a single onnx ([a40fa3a0b](https://github.com/commaai/openpilot/commit/a40fa3a0b0cfacb0010b50515bcd64aed3dbdac8)), they switched from appending the current feature to appending prev_feat and changed the feature buffer to be a 96 ring instead of a 93 ring.
modeld_v2 correctly appends prev_feat for supercombo but keeps the 93 ring and only lags by 1 frame instead of 4, so fused models get a features_buffer shifted 3 frames newer compared to openpilot.

I believe this affects the newly added models like TSFDO, RDF, etc.

all using newer models

### openpilot version

6dd3457f4fd269919cbfb95af4bd930761fbe166

### Additional info

_No response_

Quantizr:
When upstream combined the driving model into a single onnx ([a40fa3a0b](https://github.com/commaai/openpilot/commit/a40fa3a0b0cfacb0010b50515bcd64aed3dbdac8)), they changed the feature buffer to be a len 96 ring instead of a 93 ring.
modeld_v2 keeps the 93 ring and so only lags by 1 frame instead of 4, so fused models get a features_buffer shifted 3 frames newer compared to openpilot.

Fixes: #1917

Checked the older supercombo models, they were frame_skip=1 which makes the math equivalent so no regressions there.

### DIFF
--- a/openpilot/sunnypilot/modeld_v2/compile_modeld.py
+++ b/openpilot/sunnypilot/modeld_v2/compile_modeld.py
@@ -117,7 +117,8 @@ def generate_queues_and_npy(input_shapes: dict, frame_skip: int, device: str = D
   }
 
   if features_buffer:
-    queues['feat_q'] = Tensor(np.zeros((frame_skip * (features_buffer[1] - 1) + 1, features_buffer[0], features_buffer[2]),
+    feat_q_len = frame_skip * features_buffer[1] if is_supercombo else frame_skip * (features_buffer[1] - 1) + 1
+    queues['feat_q'] = Tensor(np.zeros((feat_q_len, features_buffer[0], features_buffer[2]),
                        dtype=np.float32), device=device).contiguous().realize()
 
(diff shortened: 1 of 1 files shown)


openpilot/sunnypilot/modeld_v2/compile_modeld.py
```diff
@@ -117,7 +117,8 @@ def generate_queues_and_npy(input_shapes: dict, frame_skip: int, device: str = D
   }
 
   if features_buffer:
-    queues['feat_q'] = Tensor(np.zeros((frame_skip * (features_buffer[1] - 1) + 1, features_buffer[0], features_buffer[2]),
+    feat_q_len = frame_skip * features_buffer[1] if is_supercombo else frame_skip * (features_buffer[1] - 1) + 1
```
Discountchubbs, line 120: Nice

Discountchubbs (approved):
Thanks for the PR!

PR #1926, opened 2026-08-20
[TIZI/TICI] ui: dynamic path width color

### LINKED ISSUE #1441: Path color change from green to gray on lateral disengage
### Describe the bug

Prior to the rewrite, the intended path display would transition from green and vehicle width to gray and skinny to indicate lateral disengagement whenever openpilot lateral control turned off. After the update, the only indication of control status is the colored boarder around the whole screen.

I haven't been following dev releases during the rewrite so I'm not sure if this is deliberate or what, but the previous strategy was a lot clearer for the driver (and maybe more so for a passenger who might be interested to know when the driver vs comma is in primary control). 

f5d85a9ecd10a3de/0000053a--137671eb38

### openpilot version

release-tizi

### Additional info

_No response_

granolaFPV:
### Summary of Changes
Resolves issue #1441. 

Previously, `_update_model()` hardcoded a `0.9m` half-width and `_draw_path()` tied path color blending strictly to `allowThrottle`. Because sunnypilot's MADS decouples lateral steering from longitudinal control, the on-screen driving path remained wide and green even when lateral steering was disengaged (such as in `LONG_ONLY` mode or during manual steering override).

This PR:
1. **Dynamic Path Half-Width**: Introduces dynamic half-width scaling (`0.90m` when steering is active vs `0.40m` when disengaged).
2. **Smooth Width Transitions**: Applies a `FirstOrderFilter` (`_width_filter`) to path width transitions to eliminate abrupt visual popping.
3. **Decouple Color Blending from Throttle**: Colors the path green/rainbow based on active lateral steering rather than throttle allowance.
4. **Intuitive Throttle Override Behavior**: Evaluates `mads` steering state directly so that accelerating (longitudinal override) preserves the thick colored steering path, while manual steering override / pause properly transitions to thin gray.
5. **Mode Compatibility**: Preserves expected thin-gray rendering when lateral control is off during Rainbow Mode and Experimental Mode.

### In-Vehicle Verification
- Tested in-vehicle on Comma 3X running sunnypilot.
- Verified path smoothly transitions from thick/green to thin/gray upon MADS lateral disengagement.
- Verified path smoothly returns to thick/green upon MADS re-engagement.
- Verified path stays thick/colored when overriding throttle with gas pedal.
- Verified Rainbow Path renders cleanly.

### Notes
- I have only tested on 3X and have not tested on Experimental mode.
- This is the first PR I've ever submitted to any project, and the code was generated with help from Antigravity.


sunnyhaibin (approved):
@granolaFPV Thanks for the PR! I've made a few changes:
- Pulled majority of the changes out of upstream's `ModelRenderer` into `ModelRendererSP`. For sunnypilot-related changes we try to move everything into their own modules as much as possible.
- Simplified `_lateral_active` since `ui_state.status` already handles most of them
- Width filter 0.25s -> 0.1s to match torque bar speed
- Rainbow gated on `_lateral_active`. Lateral off -> falls through to gray. Experimental keeps accel-colors always.

sunnyhaibin:
Turned out I missed the MADS only mode for overrides with path width. Will open a PR to fix :sweat_smile: 
- https://github.com/sunnypilot/sunnypilot/pull/1938

granolaFPV:
Thanks sunny for adapting and merging!

In my testing yesterday on Staging, I observed that the width dynamically scales when disabled/enabled, but the color of the line remains green even when lateral control is disabled (Comma 3X).

granolaFPV:
In case it's helpful: I looked into why the line stays green when lateral is disabled on staging.

In `_draw_path()`, `self._blend_filter.update(int(allow_throttle))` is still driving the color transition based on throttle allowance. Because `allow_throttle` is `True` when cruising (or whenever openpilot longitudinal is not controlling the vehicle), `blend_factor` remains `1.0`, keeping the path 100% `THROTTLE_COLORS` (green) even when lateral disengages.

To keep changes localized to `ModelRendererSP` in line with keeping sunnypilot logic in its own modules, here are two ideas on how we could address this — ignore if these are bad ideas :)

---

### Option 1: Minimal 1-line hook in `model_renderer.py` (Mirrors `_get_path_half_width()`)

This follows the exact same pattern already used for `self._get_path_half_width()`, keeping the upstream diff to a single delegated call:

**`openpilot/selfdrive/ui/onroad/model_renderer.py`:**
```diff
-   self._blend_filter.update(int(allow_throttle))
+   self._blend_filter.update(self._get_path_blend_target(allow_throttle))
```

**`openpilot/selfdrive/ui/sunnypilot/onroad/model_renderer.py`:**
```python
def _get_path_blend_target(self, allow_throttle: bool) -> int:
  return int(self._lateral_active)
  # Or if we want to preserve upstream throttle/braking color blending while steering is active:
  # return int(self._lateral_active and allow_throttle)
```

---

### Option 2: 100% inside `ModelRendererSP` (0 changes to `model_renderer.py`)

If the goal is to leave `openpilot/selfdrive/ui/onroad/model_renderer.py` completely untouched, `ModelRendererSP` can define `_blend_filter` as a property. Because `ModelRenderer` inherits from `ModelRendererSP`, its `self._blend_filter = ...` assignment in `__init__` is routed through the property setter, wrapping the filter:

**`openpilot/selfdrive/ui/sunnypilot/onroad/model_renderer.py`:**
```python
class LateralBlendFilter:
  def __init__(self, inner_filter, renderer_sp):
    self._filter = inner_filter
    self._sp = renderer_sp

  def update(self, allow_throttle: int) -> float:
    target = int(self._sp._lateral_active)
    # Or: target = int(self._sp._lateral_active and bool(allow_throttle))
    return self._filter.update(target)

  @property
  def x(self) -> float:
    return self._filter.x


class ModelRendererSP:
  def __init__(self):
    self.rainbow_path = RainbowPath()
    self.chevron_metrics = ChevronMetrics()
    self._width_filter = FirstOrderFilter(0.9, 0.1, 1 / gui_app.target_fps)

  @property
  def _blend_filter(self):
    return getattr(self, "_sp_blend_filter", None)

  @_blend_filter.setter
  def _blend_filter(self, filter_obj):
    self._sp_blend_filter = LateralBlendFilter(filter_obj, self)

  ...
```

PR #1937, opened 2026-08-22
ui: Update default model name after eGPU hot-plug

MarciMilligan:
## Summary
The default model name did not update in the UI when Chestnut was hot plugged while offroad.

Use the shared Chestnut readiness check so the model catalog and default model label respond consistently to offroad eGPU hot plug.

## Verification
- Ran unit tests
- Verified behavior on tizi UI
    - Connect Chestnut: model catalog and default model label update to display big models
    - Disconnect Chestnut: model catalog and default model label update to display small models

### DIFF
--- a/openpilot/common/hardware/usb.py
+++ b/openpilot/common/hardware/usb.py
@@ -9,6 +9,10 @@ TYPEC_CC_ORIENTATION_PATH = Path("/sys/class/power_supply/usb/typec_cc_orientati
 PRIMARY_USB_CONTROLLER = "a600000.ssusb"
 
 
+def is_chestnut_ready(vendor_id: int, product_id: int, product: str) -> bool:
+  return (vendor_id, product_id) in CHESTNUT_USB_IDS and product == f"custom {CHESTNUT_FW_VERSION}-CLEAN"
+
+
 def get_usb_topology() -> set[str]:
   try:
     return set(os.listdir(USB_DEVICES_PATH))
--- a/openpilot/selfdrive/modeld/helpers.py
+++ b/openpilot/selfdrive/modeld/helpers.py
@@ -7,7 +7,7 @@ import tempfile
 from pathlib import Path
 
 from openpilot.common.file_chunker import get_manifest_path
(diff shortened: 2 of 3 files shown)


nayan8teen (commented):
Hot plug/unplug works fine already. Do you have a ready to use chestnut setup?

If hot plug/unplug isn't working for you, please open up an issue with logs attached so we can investigate.

MarciMilligan:
> Hot plug/unplug works fine already. Do you have a ready to use chestnut setup?
> 
> If hot plug/unplug isn't working for you, please open up an issue with logs attached so we can investigate.

Hot plug itself is working fine, but the default model text isn't updating when you hot plug without this change.

nayan8teen:
> > Hot plug/unplug works fine already. Do you have a ready to use chestnut setup?
> > 
> > If hot plug/unplug isn't working for you, please open up an issue with logs attached so we can investigate.
> 
> Hot plug itself is working fine, but the default model text isn't updating when you hot plug without this change.

It does work in our tests. Do you mind opening an issue instead with logs so we can look into it?

sunnyhaibin:
> > Hot plug/unplug works fine already. Do you have a ready to use chestnut setup?
> > If hot plug/unplug isn't working for you, please open up an issue with logs attached so we can investigate.
> 
> Hot plug itself is working fine, but the default model text isn't updating when you hot plug without this change.

I just tested this on my C3X and it is working for me with hot unplug/plug. Are you using a comma four? Can you get some logs locally if it's not working for you?

MarciMilligan:
I'm on C3X and am still seeing `CD210 (Default)` listed after hot-plugging the Chestnut on the master branch. The model catalog updates and big models show up, but the default model name sticks. I'm wondering if there is some quirk with `ui_state.usbgpu` being set?

Are any logs emitted offroad that would even reflect the state here in the first place? I don't know where to look for these.

MarciMilligan:
I'm realizing that this issue only occurs if we boot up without the Chestnut connected (`ui_state.usbgpu_compiled == false`?). It works fine if we start with it connected.

What is the expected behavior if we boot without the Chestnut connected? Should big models be available in the first place?

sunnyhaibin:
Thanks for the PR! This is now fixed in https://github.com/sunnypilot/sunnypilot/pull/1948

Issue #1963, opened 2026-08-25
modeld_v2: eGPU load failure leaves modeld stopped with no fallback


character-s:
### Environment
- comma four + chestnut (RX 9060 XT 8GB), sunnypilot staging-chestnut
- Lexus GS 450h 2018 (LEXUS_GS_F platform)

### What happens
When `ModelState(..., usbgpu=True)` raises during load, modeld stops for the rest of the route:

1. `openpilot/sunnypilot/modeld_v2/modeld.py` decides the eGPU path with only `USBGPU = usbgpu_present()` — it never checks whether the *active bundle* is an eGPU build. A QCOM-only bundle (e.g. the default small model) selected while the chestnut is plugged in loads on `DEV='AMD'` and fails in ~4s with `tinygrad.engine.jit.JitError: args mismatch in JIT: captured=(..., 'QCOM') expected=(..., 'AMD')`.
2. The load thread has no try/except, so the exception is not recorded in cloudlog. The thread exits, `model` stays `None`, and main raises `RuntimeError("eGPU model load failed or timed out (60s)")`. This reports a timeout even when the load thread exits after about 4 seconds.
3. There is no fallback to the built-in small model. This is inconsistent with the `bigModelFailed` alert text, which says "small model is still available".
4. manager does not restart modeld, so modelV2 remains absent for the rest of the route and selfdriveState stays `disabled`.

Measured on-car: during a 27-minute route, onroadEvents messages (~200 per 60s segment) carried `processNotRunning` and `bigModelLoading` continuously, with zero modelV2, longitudinalPlan, cameraOdometry, or drivingModelData messages in the rlog.

### Repro
1. With the chestnut unplugged, select a QCOM-only model bundle (offroad).
2. Plug the chestnut in, start the car.
3. `Big Model Failed` alert; modeld_tinygrad stops and does not come back for the rest of the route.

Other load-time exceptions on the eGPU path produce the same result — see #1964 for a boot-timing case that produces one on unmodified hardware.

### Fix
Catch the load error, log it, and fall back to the pkl compiled into the build. The fallback must not reuse the active bundle's pkl — an eGPU bundle's pkl is an AMD build and `load_oob` raises when it is opened on the built-in GPU, which makes the naive fallback itself fatal. We have been running this on-car since 2026-08-25. A PR with the fix follows.

PR #1965, opened 2026-08-25
modeld_v2: retry the eGPU model load on am_usb lock contention

### LINKED ISSUE #1964: modeld_v2: /tmp/am_usb lock contention at boot makes the eGPU model load fail with no retry
### What happens
tinygrad's AMD-over-USB backend guards the device with `/tmp/am_usb:<bus>-<port>.lock`, taken with `LOCK_EX | LOCK_NB` (`tinygrad/runtime/support/system.py`) — a second process attempting to acquire the lock fails immediately. modeld tries once and gives up:

```
eGPU model load failed
eGPU model unavailable (ExceptionGroup('No interface for AMD:0 is available',
   [FileNotFoundError(2, ...), RuntimeError('no pcie'),
    RuntimeError('Failed to acquire lock file am_usb:4-2 ...')]))
```

In three observed boots with the same hardware, cable, and configuration on the same day, loads starting at 34.0s and 34.6s after boot succeeded. The load starting at 35.3s failed to acquire the lock. Because of #1963 (no fallback, no restart), this left modeld unavailable for the rest of that route.

The failure is not the 60s timeout: `t.join()` returns early because the thread raised within seconds. Rebooting clears the contention because the process holding the flock is terminated.

I could not identify the holder from the existing logs. No other openpilot code acquires this lock (`usbgpu_present()` and `common/hardware/usb.py` only read sysfs), so a previous modeld instance that has not exited yet is one possible contender.

### Fix
Retry the load when the nested exception chain contains `"Failed to acquire lock file"` (up to 5 attempts, 3s between attempts; the join timeout is extended from 60s to 72s). We have been running this on-car since 2026-08-25. A PR follows.

character-s:
tinygrad guards the AMD-over-USB device with a non-blocking flock on `/tmp/am_usb:<bus>-<port>.lock`. When another process still holds it while modeld loads the model, the load fails instantly with `Failed to acquire lock file` nested in `ExceptionGroup("No interface for AMD:0 is available")`, modeld raises and exits, and the whole drive runs without the eGPU model — even though a reboot clears it. In our logs this was a boot-time race: a load starting ~35s after boot failed, ~34s succeeded. Fixes #1964.

This is now retry-only; `ModelState` and bundle handling are untouched. The earlier fallback to the built-in model has been dropped per review — the proper fallback is being worked on by the team, and #1963 has the failure data from our runs if useful.

- Retries the load when the am_usb flock is contended: up to 5 attempts, 3s apart; the join timeout is extended accordingly (60s to 72s). Other load errors are not retried.
- Catches load-thread exceptions and includes them in the RuntimeError. Previously the thread had no try/except, so the only visible message was `eGPU model load failed or timed out (60s)` even when the load had failed within seconds.

Verification:
- `ruff check` / `python -m py_compile` on the touched files
- unit tests for the lock-contention detection and retry logic: `openpilot/sunnypilot/modeld_v2/tests/test_egpu_load.py`
- A build containing this retry has been running on-car since 2026-08-25 in a staging-chestnut based build (comma four + chestnut, Lexus GS 450h); the eGPU success path is unchanged.


openpilot/sunnypilot/modeld_v2/modeld.py
Discountchubbs, line 1: The built-in default model will not work with modeld_v2. We are working on a proper fallback.

openpilot/sunnypilot/modeld_v2/modeld.py
```diff
@@ -50,6 +50,53 @@
...
+  import glob
+  import subprocess
+  try:
+    locks = glob.glob("/tmp/am_usb:*.lock")
+    if not locks:
+      return "no lock file"
+    out = subprocess.run(["fuser", "-v"] + locks, capture_output=True, text=True, timeout=5)
+    return " ".join((out.stdout + out.stderr).split())[:300] or "nobody"
+  except Exception:
+    return "unknown"
```
Discountchubbs: Too verbose.

openpilot/sunnypilot/modeld_v2/modeld.py
```diff
@@ -101,7 +148,19 @@ def __init__(self, cam_w: int, cam_h: int, usbgpu: bool = False):
...
+    pkl_path = None if force_builtin_pkl else _find_driving_pkl(model_bundle)
+    if pkl_path is None and force_builtin_pkl:
+      fallback = str(modeld_pkl_path(usbgpu=False))
+      if _pkl_exists(fallback):
+        cloudlog.warning(f"loading the built-in pkl {fallback} instead of the active bundle")
+        pkl_path, model_bundle = fallback, None
+        # the built-in pkl is not the active bundle, so the bundle's generation and smoothing
+        # overrides do not describe it - keep them consistent with model_bundle=None
+        self.generation = None
+        self.LAT_SMOOTH_SECONDS = self.LONG_SMOOTH_SECONDS = 0.0
```
Discountchubbs: You still need an active bundle. fallback to small model from our model manager as default will not work.

Discountchubbs (changes requested):
The built-in default model will not work with modeld_v2. We are working on a proper fallback.

Discountchubbs:
This entire PR is LLM-generated. Please explain the changes in your own words.

https://github.com/sunnypilot/sunnypilot/blob/master/docs/AI_POLICY.md

character-s:
Thank you for the review. I am a Japanese speaker — my English reading is fine but I am not a native speaker, so I used a translator to write this reply.

In my own words: on Aug 24 and 25 this happened on my own car (comma four + chestnut). About 5 times on the 24th and twice on the 25th, even with the GPU connected, the screen flashed white and then went to the error. I read the logs myself and identified the causes. I put the fixes on my own comma 4 and it has been running fine since. Sometimes it worked after rebooting a few times, so I figured this was probably happening to others too, and that is why I decided to make this PR.

It is true that I used AI, but it was to help me with the submission — the analysis and the testing on my car are my own work.

About the review: since you are working on a proper fallback, I will drop the fallback part and reduce this PR to the lock-contention retry only.

Discountchubbs:
> Thank you for the review. I am a Japanese speaker — my English reading is fine but I am not a native speaker, so I used a translator to write this reply.
> 
> In my own words: on Aug 24 and 25 this happened on my own car (comma four + chestnut). About 5 times on the 24th and twice on the 25th, even with the GPU connected, the screen flashed white and then went to the error. I read the logs myself and identified the causes. I put the fixes on my own comma 4 and it has been running fine since. Sometimes it worked after rebooting a few times, so I figured this was probably happening to others too, and that is why I decided to make this PR.
> 
> It is true that I used AI, but it was to help me with the submission — the analysis and the testing on my car are my own work.
> 
> About the review: since you are working on a proper fallback, I will drop the fallback part and reduce this PR to the lock-contention retry only.

Hello there. there has got to be a better way to do this. I will keep your PR and issue open. The fallback PR is now made, changing gpu locks may be better to do somewhere else and import to modeld for clean abstraction.

character-s:
I moved the retry code to a separate file (sunnypilot/modeld_v2/egpu_load.py). Now modeld just imports it. The behavior is the same. Thank you for keeping the PR and issue open.

character-s:
For the record: the contention here has been traced to the import-time `Device.DEFAULT` probe in `modeld_v2/compile_modeld.py` taking the flock against itself (PeterPhuTran's comment in #1964, fixed by #2033).

My own device agrees. Reading `/proc/locks` right before the load, 71 of 71 starts already hold that flock themselves — so this retry is waiting on a holder that can never exit. Every one of the 66 processes that hit the lock with the retry active failed anyway.

That leaves only the case where a genuinely different process holds it, which has not been reproduced on a car. Happy to close once #2033 lands; leaving it open in case the maintainers want the retry as a safety net for that case.

PR #1974, opened 2026-08-28
modeld_v2: big to small model fallback

### LINKED ISSUE #1963: modeld_v2: eGPU load failure leaves modeld stopped with no fallback

Discountchubbs:


Discountchubbs:
https://github.com/commaai/openpilot/pull/38704

https://github.com/commaai/openpilot/pull/38684

https://github.com/commaai/openpilot/issues/38632

https://github.com/commaai/openpilot/issues/38505



next PR will need to look into these openpilot issues.

sunnyhaibin (approved):
LGTM

PR #1993, opened 2026-09-05
modeld_v2: realize frames on npy -> amd

Discountchubbs:
Copy frames from QCOM to CPU to chestnut/AMD. Realized frames on chestnut results in frozen frames since visionipc runs on qcom not amd. This was not an issue on older chestnut models because we warped in qcom, so the frames were always available in shared memory via zero-copy. New tinygrad models require the same device, so to warp on amd, the input frames need to be npy tensors. As a result of this, we had to transition to a monolithic compile structure on modeld to match upstream. The resulting files in modeld and models directory were refactored/removed due to newly deprecated code.


nayan8teen (dismissed):

Discountchubbs:
I'm feeling like a major red diff is about to occur.

nayan8teen (approved):
Double Approval

PR #1995, opened 2026-09-06
models: preserve verified chunks and retry interrupted downloads

dustin-decker:
When a model transfer times out, `_process_artifact` deletes every chunk for that artifact, including chunks that already downloaded successfully. The request loop then clears the failed selection, so the UI quickly loses the error and the next attempt starts over.

Preserve files on failure and reuse only chunks that pass the existing SHA-256 checks. Retry transient connection/body errors and HTTP 408/429/500/502/503/504 up to four total attempts, with cancellable 2/4/8-second backoff. Permanent errors fail immediately. Keep the failed selection visible until another download, and restore a missing manifest for fully verified cached chunks without rewriting an existing manifest. Activation still happens only after successful verification.

Validation:
- 24 downloader/import tests passed against this upstream manager and test file using a temporary local harness for unavailable compiled dependencies and the surrounding BluePilot compatibility modules. This is not a full native upstream test run.
- Added real HTTP regression coverage for truncated bodies, preserving completed chunks after retries are exhausted, immediate 404 failure, cancellation during backoff, persistent error status, and missing/existing manifests.
- Ruff and `git diff --check` pass.
- The equivalent implementation on BluePilot passed 60 tests on a comma device with its real extensions, plus `scons -j4`. GWM v8 downloaded, verified, and became selected. During a subsequent favorites cache fill, a DNS failure recovered automatically and a mid-transfer read timeout resumed from the verified 50% boundary.

Implementation assisted by OpenAI Codex; adversarial review of the equivalent BluePilot change by Claude Fable 5.1. No driving-control changes or driving-inference validation are included.

### DIFF
--- a/openpilot/sunnypilot/models/manager.py
+++ b/openpilot/sunnypilot/models/manager.py
@@ -154,6 +154,41 @@ class ModelManagerSP:
     del self._download_start_times[artifact.fileName]
 
   async def _process_artifact(self, artifact, destination_path: str) -> None:
+    # retry transient failures, including errors while streaming the body.
+    # Rechecking hashes on each attempt preserves complete chunks without trusting partials.
+    try:
+      for attempt in range(4):
+        try:
+          await self._process_artifact_once(artifact, destination_path)
+          return
+        except (requests.ConnectionError, requests.Timeout, requests.exceptions.ChunkedEncodingError, requests.HTTPError) as e:
+          if isinstance(e, requests.HTTPError) and (e.response is None or e.response.status_code not in (408, 429, 500, 502, 503, 504)):
+            raise
+          if attempt == 3:
+            raise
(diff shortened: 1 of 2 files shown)


AmyJeanes:
Hey, I've been rewriting the model downloader to use clientside chunking anyway which already does a lot of what is in this PR and I've pulled in the remaining bits into my PR too: https://github.com/sunnypilot/sunnypilot/pull/1998

Note still WIP but just wanted to let you know

dustin-decker:
@AmyJeanes fantastic, looking forward to faster downloads in the future! Definitely will be important with the big models.

AmyJeanes:
If your PR goes in first mine applies nicely on top of yours now so regardless we'll get your improvements in!

PR #1997, opened 2026-09-06
Use patched opendbc for Tesla Juniper

sarumpj-arch:


TheSecurityDev:
You need to open the pr against your own fork's repo if you're trying to make the change for yourself.

PR #1998, opened 2026-09-06
models: whole-file model downloads with parallel byte ranges

AmyJeanes:
## What

Models are published to HuggingFace as one whole file and the model manager downloads them with 12 parallel HTTP byte-range requests, replacing the `.chunkNNofMM` + `.chunkmanifest` split that the client fetched one chunk at a time over a single connection.

- **Parallel ranges, written in place.** The file is created at its final size and each 8 MB piece streams straight to its offset from its own worker thread, so a model is never copied or assembled and only ever occupies its own size on disk.
- **Resumable and power-loss safe.** A `<file>.download` sidecar records which pieces have landed. A piece is only recorded after `fdatasync`, and the sidecar itself is written via fsync + rename, so pulling the plug mid-download (engine crank, battery disconnect) can never leave resume state naming bytes that did not reach flash. Cancel and transport failures keep the file + sidecar; the next selection resumes with "Resuming interrupted download". The sidecar is keyed to the artifact's sha256, size and piece layout, so a republished model never resumes from old bytes.
- **Retries and fallbacks.** Transient errors (connection, timeout, truncated body, HTTP 408/429/500/502/503/504) retry a piece up to 3 times. If a piece still fails, the whole transfer is retried up to 4 times with a cancellable 2/4/8 s backoff, each time resuming from the sidecar, so nothing on disk is fetched twice. Any other HTTP error fails at once. A failed download stays on the Models row until the next request or a cancel, instead of vanishing on the next tick. A server that answers the range probe with 200 gets one plain stream instead.
- **Verification.** The whole file is hashed after download; on mismatch the file and sidecar are deleted. Legacy `.chunk*` files for the same name are purged first, because `open_file_chunked` prefers a chunk manifest over a whole file of the same name.
- **UI.** `DownloadProgress.speed` (bytes/s, smoothed) is appended to the capnp. `Model.Type` gains `driving`, which new builds emit for the whole pkl instead of `chunked`; `chunked` stays in the enum so an upgrading device can still parse its cached v22 catalog. The big UI's Models page shows `12.3 MB/s` next to the progress bar and the mici models panel shows it beside the percentage. A failed download stays on the row until the next request (see #1995 below), so the mici error state names the failed model instead of animating dots. The mici panel also gains a clear-cache tile (cache size, slide to confirm) matching the big UI's Clear Model Cache row, since it had no way to drop cached models. ETA is derived from the measured rate. Progress is published every 0.5 s instead of every 128 KB.

## Why

HuggingFace's Xet CAS throttles each connection to roughly 1-2 MB/s (erratically, dips to 300 kB/s) but does not limit connection count. The serial client was link-idle most of the time.

Measured on a comma 3X on the same WiFi, sunnypilot dataset files:

| | throughput |
|---|---|
| current client, single stream | ~2 MB/s |
| 12 range connections | 17-24 MB/s steady, 30-60 MB/s peak |

A 1.7 GB big model goes from ~14 min to 1-2 min. Disk is not the bottleneck (eMMC writes at 56-67 MB/s with fdatasync).

## Rollout (maintainers)

1. **Selector version is now 20** and the client reads `driving_models_v23.json` / `driving_models_chestnut_v26.json`. Nothing reads those files until this merges and nothing on master changes until they exist, so the order is: migrate the catalog first, merge second. Clients on 19 keep reading v22 / chestnut_v25 untouched.
2. **Migrate the catalog, no recompile.** `release/ci/migrate_whole_file_models.py` (dispatch: `migrate-whole-file-models`, plain runner, uses the HF token and docs deploy key already in the repo) downloads each model's chunks, joins them, checks the result against the `download_uri.sha256` the manifest already carries (it has always been the whole-file hash), uploads the whole pkl beside the chunks, confirms HuggingFace's stored LFS hash, and writes v23 / chestnut_v26 as copies of v22 / chestnut_v25 with the chunk lists dropped, `type: driving` and selector 20. v22 and chestnut_v25 pin the same tinygrad ref as master (`e837e367`), so the migrated files are byte-identical to what selector-19 clients run today. A rerun skips models already stored with the right hash, and the manifest is only written once every model migrated, so it never lists a model that cannot be downloaded. Run it once for qcom (22 → 23) and once for chestnut (25 → 26), roughly 24 GB each way in total, then merge. Rollback is a revert of this PR: v22 and its chunks are never touched. The script and its workflow are one-offs to delete afterwards.
3. **Publishing new models** uses the existing `build-single-tinygrad-model` workflow with `json_version=23` (or 26 for chestnut) and `version=20`. `sunnypilot-build-model` now copies only the compiled `*_tinygrad.pkl` into the artifact, and `compile_modeld.py` no longer chunks it. `json_parser.py` in the docs repo needs no change: it copies each model entry from the build's `metadata.json` and only rewrites the URL, so a whole-file build already comes out with a plain `download_uri`. The fork manifests below were produced by the parser as it is.
4. **A full recompile** (`build-all-tinygrad-models`) still works and stays the normal path for the next bump (21 / v24). It copies the previous json first and stamps every bundle with `set_min_version` on each publish, so partway through a run the new manifest lists bundles whose whole file does not exist yet; that is why this bump uses the migration, which writes a complete manifest in one go. If build-all is used instead, do not pre-create v23: it creates the next number itself.
5. **Upgrading users** get their selected slot reset to default (selector mismatch, as with every bump) and re-download from the new manifest.
6. **Default models are untouched.** `build-default-models`, `sunnypilot-build-prebuilt`, `download-hf-model-chunks` and `release/ci/model_generator.py` still ship the defaults chunked inside the release, and `open_file_chunked` keeps reading both forms.
7. `build-single-tinygrad-model` and `build-all-tinygrad-models` gained a `docs_repo` input (default `sunnypilot/sunnypilot-models`), and build-all's setup job now checks out the repo and branch it was dispatched from rather than `sunnypilot/sunnypilot`'s default branch. That keeps the manifest's `tinygrad_ref` tied to the code that compiled the models, and lets a fork run the full rebuild against its own dataset and docs repo. The live manifest test skips entries still at the previous selector version, since a build-all manifest starts as a copy of the previous one.

## Relationship to #1995

#1995 adds per-artifact retries, keeps verified chunks on failure, keeps the failed selection visible, and restores a missing chunk manifest. This PR carries the same retry semantics (same transient status set, 4 attempts, 2/4/8 s cancellable backoff) and the same sticky failed selection, with tests for both. Keeping partial data is covered by the sidecar resume, and there are no chunk manifests any more, so those two parts do not apply. If #1995 lands first the rebase is mechanical: this PR's version of the artifact path and test file, plus #1995's change to the request loop, which is already included here.

## Testing

- `openpilot/sunnypilot/models/tests/test_manager_download.py` drives the real downloader against a local threaded range server: piece tiling, actual parallelism (a stalled piece does not block the others), retry, 4xx, cancel via `ModelManager_DownloadRef`, per-transfer retry resuming from the sidecar, 503 gives up after every attempt, 404 not retried, cancel during backoff, failed/finished/cancelled row state, resume (skips recorded pieces; ignores a sidecar for another size, another hash, a truncated file, or malformed JSON), fdatasync/fsync accounting, hash mismatch cleanup, legacy chunk purge, cache clearing, and the no-range fallback.
- The mici models panel was rendered off-device (hidden raylib window, synthetic `modelManagerSP` messages) in the idle, downloading, verifying, failed and downloaded states. It reads only bundle and artifact status plus progress, so it needs nothing beyond the additive `speed` field.
- `openpilot/sunnypilot/models/tests/test_migrate_whole_file_models.py` drives the migration script against a local range server and a fake HuggingFace API: join, upload and entry rewrite; a rerun keeps stored files; a stale stored file is replaced; a whole-file hash mismatch stops before any upload; a corrupt chunk is retried and then fails; dry run; `only`/`limit` selection; unchunked entries; output in the docs parser's style.
- **The catalog migration was rehearsed end to end on a fork.** `migrate_whole_file_models.py` read the chunks named by sunnypilot's live v22 and chestnut_v25, joined and verified all 89 models against their manifest hashes, published the whole files to a fork dataset and wrote v23 / chestnut_v26 to a fork docs repo, through the `migrate-whole-file-models` workflow on a GitHub-hosted runner: the 12 chestnut models (17 GB) took 8 minutes, and a second qcom run recognised all 77 files as already stored and only wrote the manifest. The migrated manifests differ from their sources in nothing but the selector version, the model type, the dropped chunk lists and the dataset name.
- On a comma 3X reading those manifests, the real model manager downloaded the migrated BMv6 (787 MB in 33 s, ~36 MB/s) and CD210 (84 MB in 5 s), each replacing a cached file of the same name with a different hash, verified against the manifest with no sidecar or legacy chunk files left. Both hashes equal the entries in sunnypilot's current chunked manifests, so users get byte-identical models.
- `RUN_INTEGRATION_TESTS=1` adds a live check that every artifact in both manifests answers a range request with 206.
- End to end on a comma 3X against a fork of the HF dataset and docs repo: three bundles (two qcom, one chestnut) compiled with `build-single-tinygrad-model` on a self-hosted device runner, then a full `build-all-tinygrad-models` run (qcom, `set_min_version=20`, fork `hf_repo` and `docs_repo`) that created the next manifest from the previous one, rebuilt every bundle in it and published whole files that answer range requests, downloaded through the Models page, cancelled mid-transfer and resumed, hash verified against the manifest.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

https://claude.ai/code/session_01EGMnVnSk5inGTrd7kuDVG9

### DIFF
--- a/.github/workflows/build-all-tinygrad-models.yaml
+++ b/.github/workflows/build-all-tinygrad-models.yaml
@@ -20,6 +20,11 @@ on:
         required: false
         type: string
         default: 'sunnypilot/sunnypilot_models_v1'
+      docs_repo:
+        description: 'GitHub repo holding the driving_models JSON on its gh-pages branch'
+        required: false
+        type: string
+        default: 'sunnypilot/sunnypilot-models'
 
 jobs:
   setup:
@@ -31,10 +36,11 @@ jobs:
       model_matrix: ${{ steps.set-matrix.outputs.model_matrix }}
       tinygrad_ref: ${{ steps.get-tinygrad-ref.outputs.tinygrad_ref }}
     steps:
+      # the repo and branch this run was dispatched from: the same code build_model compiles with,
+      # so the manifest's tinygrad_ref always matches the models in it (and forks build their own)
       - name: Checkout sunnypilot repo
         uses: actions/checkout@v4
         with:
-          repository: sunnypilot/sunnypilot
           path: sunnypilot
           submodules: recursive
 
@@ -47,10 +53,10 @@ jobs:
           echo "tinygrad_ref=$ref" >> $GITHUB_OUTPUT
           echo "tinygrad_ref is $ref"
 
-      - name: Checkout docs repo (sunnypilot-models, gh-pages)
+      - name: Checkout docs repo (gh-pages)
         uses: actions/checkout@v4
         with:
-          repository: sunnypilot/sunnypilot-models
+          repository: ${{ inputs.docs_repo }}
           ref: gh-pages
           path: docs
           ssh-key: ${{ secrets.CI_SUNNYPILOT_DOCS_PRIVATE_KEY }}
@@ -118,6 +124,7 @@ jobs:
       json_version: ${{ needs.setup.outputs.json_version }}
       target_hardware: ${{ github.event.inputs.target_hardware }}
       hf_repo: ${{ github.event.inputs.hf_repo }}
+      docs_repo: ${{ inputs.docs_repo }}
       set_min_version: ${{ github.event.inputs.set_min_version }}
       tinygrad_ref: ${{ needs.setup.outputs.tinygrad_ref }}
     secrets: inherit
@@ -162,6 +169,7 @@ jobs:
       target_hardware: ${{ github.event.inputs.target_hardware }}
       artifact_suffix: -retry
       hf_repo: ${{ github.event.inputs.hf_repo }}
+      docs_repo: ${{ inputs.docs_repo }}
       set_min_version: ${{ github.event.inputs.set_min_version }}
       tinygrad_ref: ${{ needs.setup.outputs.tinygrad_ref }}
     secrets: inherit
--- a/.github/workflows/build-single-tinygrad-model.yaml
+++ b/.github/workflows/build-single-tinygrad-model.yaml
@@ -39,6 +39,11 @@ on:
         required: false
         type: string
         default: 'sunnypilot/sunnypilot_models_v1'
+      docs_repo:
+        description: 'GitHub repo holding the driving_models JSON on its gh-pages branch'
+        required: false
+        type: string
+        default: 'sunnypilot/sunnypilot-models'
       set_min_version:
         description: 'Minimum selector version'
         required: false
@@ -107,6 +112,11 @@ on:
         required: false
         type: string
         default: 'sunnypilot/sunnypilot_models_v1'
+      docs_repo:
+        description: 'GitHub repo holding the driving_models JSON on its gh-pages branch'
+        required: false
+        type: string
+        default: 'sunnypilot/sunnypilot-models'
 env:
   RECOMPILED_DIR: recompiled${{ inputs.recompiled_dir }}
   JSON_FILE: docs/docs/driving_models_${{ inputs.target_hardware == 'chestnut' && 'chestnut_v' || 'v' }}${{ inputs.json_version }}.json
@@ -136,7 +146,7 @@ jobs:
       - name: Checkout docs repo
         uses: actions/checkout@v4
         with:
-          repository: sunnypilot/sunnypilot-models
+          repository: ${{ inputs.docs_repo || 'sunnypilot/sunnypilot-models' }}
           ref: gh-pages
           path: docs
           ssh-key: ${{ secrets.CI_SUNNYPILOT_DOCS_PRIVATE_KEY }}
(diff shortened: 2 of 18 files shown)


Discountchubbs:
Hey there! Whenever this is ready for review please note, both @sunnyhaibin and I will ask you to split this pr up into multiple small <500loc PRs so we can review it effectively.

AmyJeanes:
Split into eight PRs as asked, each under or near 500 lines and mergeable on its own. Their merged result is byte-identical to this branch.

- #2008 mici clear cache tile (independent)
- #2009 workflows docs_repo input (independent)
- #2010 catalog migration tool (independent; run before #2015)
- #2011 parallel byte-range downloader (chunked path untouched)
- #2012 download speed, on #2011
- #2013 resumable sidecar, on #2012
- #2014 transfer retries and sticky failed row, on #2013
- #2015 the switch to selector 20, on #2014, after #2010 has been run

Closing this one in favour of them.

PR #2000, opened 2026-09-07
selfdrived: chime Big Model Ready on a real big frame, gate localizer alerts on seen

zephleggett:
Fix misleading startup alerts, the big-model ready chime, and unnecessary model manifest writes.

- **Startup alerts:** Wait for the first localization and vehicle-parameter messages before checking their status. Previously, missing messages could trigger “Posenet Speed Invalid” and temporary errors before those processes reported any results.
- **Big Model Ready:** Play the chime when the first valid big-model frame arrives. A failed or timed-out load no longer plays “Big Model Ready” after “Big Model Failed” while the small model is running.
- **Model manifests:** Repair chunk manifests only when matching chunks exist on disk and the recorded count needs updating. This avoids creating manifests for undownloaded models or repeatedly overwriting them when catalogs list different chunk counts.

### Verification

- Regression tests cover startup alerts, successful and failed big-model loads, chime rearming, and manifest repair across catalogs.
- `pytest openpilot/selfdrive/selfdrived/tests openpilot/sunnypilot/models/tests/test_manager_download.py`: 72 passed, 1 skipped.
- Ruff passed on changed files.

### DIFF
--- a/openpilot/selfdrive/selfdrived/selfdrived.py
+++ b/openpilot/selfdrive/selfdrived/selfdrived.py
@@ -85,6 +85,7 @@ class SelfdriveD(CruiseHelper):
     self.big_model_loading = False
     self.big_model_active = False
     self.big_model_failed = False
+    self.big_model_running = False
     self.big_model_ready_t = 0.
 
     # Setup sockets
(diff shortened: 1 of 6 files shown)


zephleggett:
@sunnyhaibin @Discountchubbs these are generic fixes for chestnut I found developing _accelerators_.

zephleggett:
Closing because I should probably open this upstream in openpilot.

PR #2001, opened 2026-09-07
accelerators: run the big driving model on an attached Jetson (jetlink)

zephleggett:
Add Jetlink support to run large driving models on an NVIDIA Jetson connected to the comma over USB. Users can select a model in sunnypilot, prepare it while parked, and use the local small model while the Jetson starts or reconnects.

Depends on #2000 for startup-alert fixes and the “Big Model Ready” chime on the first valid big-model frame.

### Features

- **Model settings:** Adds an **Accelerator Link** toggle and **Accelerator Model** picker in both UI layouts. Link and model changes are available only offroad. Jetlink is disabled by default.
- **Model preparation:** Downloads the selected ONNX model, transfers it to the Jetson, and builds a TensorRT engine with progress and estimated build time. Cached engines are reused and checked when the Jetson reconnects.
- **Model choice:** Uses a separate ONNX catalog for the Jetson, with Cinque Terre as the default. Existing QCOM model bundles cannot run under TensorRT.
- **Startup and recovery:** Starts on the default small model and connects in the background. A prepared Jetson can join later in the drive once controls are disengaged.
- **Power management:** Releases the USB connection when parked and idle so a configured Jetson server can suspend. Presents the connection again when needed and requests Jetson shutdown when the comma powers off, with a bounded wait.

### These are what I have tested for reliability 

| Situation | Behavior |
| --- | --- |
| Preparing a model while parked | Model settings show download/build progress and readiness. |
| Jetson is still booting | The default small model runs while Jetlink connects. |
| Large model becomes available while engaged | “Big Model Available” prompts the driver to disengage to switch. |
| Large model starts running | “Big Model Ready” chimes after the first valid big-model frame. |
| Link fails while engaged | Falls back to the small model and triggers a soft disable. “Big Model Lost” explains that the small model is running and the connection will be retried. |
| Link fails while disengaged | Falls back silently and retries. Switching back to the large model again requires disengagement. |
| Jetlink cannot initialize | An offroad accelerator-unavailable alert reports the setup error. |

While Jetlink is enabled, sunnypilot uses stock modeld and the default small model for startup and fallback. The UI reports that model rather than the stored custom small-model selection. Turning the link off restores normal model selection.

### Setup

Requires a Jetson running the matching [Jetlink server](https://github.com/zoompilot/jetlink), a separate Jetson power supply, and a USB 3 data cable from **Jetson USB-A to comma USB-C**. The reference setup is an Orin Nano Super 8 GB with JetPack 6.1 and TensorRT 10.3.

The client ships as the pinned `jetlink_repo` submodule. With the server configured and connected, enable **Accelerator Link**, select an **Accelerator Model**, and complete preparation while parked. Installing the package alone does not enable the link. Wired TCP is also available through `JetlinkEndpoint` for bench use.

### Implementation

The comma retains camera processing, image warping, output parsing, and vehicle control. The Jetson runs inference and maintains model history buffers; it has no CAN access. The comma is the USB FunctionFS gadget and the Jetson is the host.

- Adds a small accelerator module under `openpilot/sunnypilot/accelerators`. The native chestnut path retains priority and its existing implementation.
- Switches models only with fresh, valid state confirming that controls are disengaged. Inference failures reset the small-model history and rerun the frame locally; reconnect delays increase from 5 to 60 seconds. The inference timeout is 0.5 seconds.
- Keeps connection work off the model frame thread. `jetlinkd` owns USB offroad; `modeld` owns it onroad. VM tuning is applied only when enabled and restored when disabled.
- Adds accelerator state to `ModelDataV2SP` without changing the meaning of `deviceState.chestnutPresent` or `chestnutState`. Jetson telemetry is recorded in swaglog.

### Verification

- Local regression suites: **501 passed, 3 skipped** across accelerators, models, modeld_v2, selfdrived, sunnypilot selfdrived, and UI.
- Tests cover model switching, reconnects, soft-disable persistence, opt-in behavior, model selection, and native chestnut equivalence.
- Ruff passed on changed files; affected runtime modules import on a PC.
- `validate_sp_cereal_upstream.py`: **cereal compat OK**.
- On-device opt-out check: with `JetlinkEnabled` absent, setup leaves the USB gadget inactive and accelerator shutdown returns in under 1 ms.

Live camera/modeld inference against a Jetson and shutdown with an unresponsive Jetson are not exercised in CI. Sustained latency, thermal behavior, USB faults, and suspend/wake require hardware validation; the software test results do not establish driving readiness.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -18,6 +18,7 @@ bin/
 # created at launch for comma hardware PYTHONPATH (PC uses editable installs via pyproject.toml)
 /msgq
 /opendbc
+/jetlink
 /rednose
 /teleoprtc
 /tinygrad
--- a/.gitmodules
+++ b/.gitmodules
@@ -19,3 +19,6 @@
 [submodule "sunnypilot/neural_network_data"]
 	path = openpilot/sunnypilot/neural_network_data
   url = https://github.com/sunnypilot/neural-network-data.git
+[submodule "jetlink"]
+  path = jetlink_repo
+  url = https://github.com/zoompilot/jetlink.git
--- a/launch_chffrplus.sh
+++ b/launch_chffrplus.sh
@@ -79,6 +79,9 @@ function launch {
   ln -sfn rednose_repo/rednose rednose
   ln -sfn teleoprtc_repo/teleoprtc teleoprtc
   ln -sfn tinygrad_repo/tinygrad tinygrad
+  ln -sfn jetlink_repo/jetlink jetlink
+  # accelerator backends: USB gadgets, symlinks, anything needing root at boot
+  ./openpilot/sunnypilot/accelerators/setup.sh
 
   # hardware specific init
   if [ -f /AGNOS ]; then
--- a/openpilot/cereal/custom.capnp
+++ b/openpilot/cereal/custom.capnp
@@ -354,6 +354,8 @@ struct OnroadEventSP @0xda96579883444c35 {
     e2eChime @23;
     laneChangeRoadEdge @24;
     bigModelReady @25;
+    bigModelAvailable @26;
+    bigModelLinkLost @27;
   }
 }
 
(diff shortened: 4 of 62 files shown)


Discountchubbs:
Hey there! Whenever this is ready for review please note, both @sunnyhaibin  and I will ask you to split this pr up into multiple small <500loc PRs so we can review it effectively.

PR #2002, opened 2026-09-07
modeld: compile every driver camera into the prebuilt

zephleggett:
### The Issue

_maybe this https://github.com/commaai/openpilot/issues/38762_
A prebuilt release only works on the kind of device it was built on. The release build compiles the driving model and driver monitoring warp jits for the driver camera of the machine running `scons`, and nothing else. The same tree is then pushed to both the tizi and mici release branches, so one of them is missing the jits for its own camera.

The result on the wrong device is `process not running: dmonitoringmodeld` from the first boot. dmonitoringmodeld crashes before its first frame because the warp file for its camera resolution does not exist, and modeld would fail the same way on the stock model.

We hit this on our fork: our releases are cut on a comma four and every comma 3X that installed them lost driver monitoring. Cutting on a tizi, as sunnypilot does, breaks it for the comma four instead.

### The Fix
* The release build now compiles the jits for both driver cameras. tizi and mici share the same GPU, so the compiled kernels run on either and only the camera resolution differs. A source build on a device is unchanged and still compiles just its own camera.
* The chunk size estimate for the driving pkl accounts for the extra camera, so a build that comes up short fails in CI rather than on the device.
* When a device does have a build without its camera, modeld and dmonitoringmodeld now fail with a message naming the camera and what the build contains, instead of a bare `FileNotFoundError` or `KeyError`.

### Notes
* Verified on our fork: a 3X on a comma four cut release now runs driver monitoring.
* Compiling both cameras adds a few minutes to the release build.
* Downloaded model bundles are unaffected; this only concerns the pkls built by `scons`.

### DIFF
--- a/openpilot/selfdrive/modeld/SConscript
+++ b/openpilot/selfdrive/modeld/SConscript
@@ -18,15 +18,19 @@ tinygrad_root = env.Dir("#").abspath
 tinygrad_files = ["#"+x for x in glob.glob(env.Dir("#tinygrad_repo").relpath + "/**", recursive=True, root_dir=tinygrad_root)
                   if 'pycache' not in x and os.path.isfile(os.path.join(tinygrad_root, x))]
 
-def estimate_pickle_max_size(onnx_size):
+def estimate_pickle_max_size(onnx_size, num_cameras=1):
   # QCOM programs for models with spatial recurrent features can approach 2x
-  # the ONNX size. Overestimating only adds an empty trailing chunk.
-  return 2.0 * onnx_size + 10 * 1024 * 1024
+  # the ONNX size, once per camera. Overestimating only adds an empty trailing chunk.
+  return 2.0 * onnx_size * num_cameras + 10 * 1024 * 1024
 
 if arch == 'comma_arm64':
   from openpilot.common.hardware import HARDWARE
-  camera = _os_fisheye if HARDWARE.get_device_type() == "mici" else _ar_ox_fisheye
-  camera_configs = [(camera.width, camera.height)]
+  # tizi and mici share the GPU, so a prebuilt can carry the jits for both driver cameras
+  if os.getenv('PREBUILT_ALL_CAMERAS'):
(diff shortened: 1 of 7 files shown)


sunnyhaibin (commented):
Thanks for the PR! Please see the points I have raised in https://github.com/sunnypilot/sunnypilot/pull/1999#pullrequestreview-5136485858, which would also apply in this PR.

zephleggett:
Closing this: sunnypilot is already covered, from two directions.

The prebuilt never compiles the jits. `sunnypilot-build-prebuilt.yaml` builds with `SKIP_TINYGRAD_COMPILE=1 scons` and the publish job downloads every pkl from HF into `openpilot/selfdrive/modeld/models`. #1989 added both `dm_warp_*` files to that download, and the default driving pkl is compiled in `build-default-models.yaml` with both camera resolutions already, so a sunnypilot prebuilt carries both cameras on both paths. Nothing to fix here.

The gap only exists where a release is cut by actually running `scons`, which is commaai/openpilot#38762. Upstream fixed it in commaai/openpilot#38767 (`9946c521f7`, camera_configs restored to both resolutions for all platforms), so it arrives with the next sync rather than as a fork patch.

That leaves only the clearer error messages, which are in stock `modeld/*` and, per your note on #1999, not worth the sync conflicts now that the root cause is fixed at the source. Dropping them.

One small thing upstream's fix does not cover, in case it is worth a follow-up there: `estimate_pickle_max_size` still sizes the driving pkl from the ONNX size alone, while the pkl now holds a `run_model` jit per camera.

zephleggett:
> Thanks for the PR! Please see the points I have raised in [#1999 (review)](https://github.com/sunnypilot/sunnypilot/pull/1999#pullrequestreview-5136485858), which would also apply in this PR.

@sunnyhaibin Thank you for looking at this! It appears you have handled the issue in https://github.com/sunnypilot/sunnypilot/pull/1989

PR #2006, opened 2026-09-08
models: test tinygrad concurrency

Discountchubbs:
The serialized tensor weights, biases, and normalization constants reside in out-of-band binary buffers (pickle.PickleBuffer).
Both stock openpilot and sunnypilot read these buffers directly from disk into hardware memory using (struct.unpack('<q', h)[0]) and stream copying (f.readinto(pb)).
The underlying tensor buffers loaded into Qualcomm Adreno / AMD GPU memory are bit-for-bit identical.
Kernel Execution Immutability in Tinygrad and the execution graph, memory strides, floating-point weights, and kernel launch parameters remain frozen.


nayan8teen (approved):
LFGTM

PR #2018, opened 2026-09-10
mici: add a refresh models button to the models panel

AmyJeanes:
Added a refresh models button to the mici (comma four) UI, which was missing there, and also some general UX improvements around it.

In both mici and tizi UIs the refresh button will now:
- Be disabled while on-road, [as model manager is not running](https://github.com/sunnypilot/sunnypilot/blob/6135084c941d4d947dd90c78326a557c3c857f89/openpilot/system/manager/process_config.py#L171) so pressing it will essentially queue it till offroad anyway
- Be disabled while downloading other models, as the [manifest fetch happens once per manager tick](https://github.com/sunnypilot/sunnypilot/blob/6135084c941d4d947dd90c78326a557c3c857f89/openpilot/sunnypilot/models/manager.py#L326) and the [blocking download loop below it](https://github.com/sunnypilot/sunnypilot/blob/6135084c941d4d947dd90c78326a557c3c857f89/openpilot/sunnypilot/models/manager.py#L337) holds up that tick until the download finishes
- Be disabled while a refresh is queued/in progress (note: has a timeout of 20s in case it fails to refresh for some reason)

In mici UI specifically:
- The button shows "fetching..." while refreshing

In tizi UI specifically:
- The button itself changes to say "FETCHING..." while refreshing (and is also disabled), replacing the previous fire-and-forget "Fetching Latest Models" popup

Videos:

mici ui:

https://github.com/user-attachments/assets/7844e166-5148-4c24-95f7-459783783abe

tizi ui:

https://github.com/user-attachments/assets/30a0d6a4-ef5a-452c-9606-6a082c052e03

The only real question on this is what we should do in the event of a timeout. We don't know currently what failed or if it's stalled we just see that the parameter for refresh time had not been updated so we re-enable the button to try again after 20 seconds. We could have it say "failed" or something in that case instead? It would require changing the internals of the refresh though as I believe it just retries constantly if you hit refresh until it succeeds currently. Let me know your thoughts.


Discountchubbs:
Fix merge conflicts, then we can approve

Discountchubbs (approved):
LGTM

Issue #2019, opened 2026-09-10
NNLC + big/chestnut models: torque command saturates ±100% below ~6 m/s (low-speed steering limit cycle); NNLC off fixes it — with data


PeterPhuTran:
### Describe the bug

**NNLC (Neural Network Lateral Control) combined with the big / chestnut driving models drives the Toyota torque controller into a ±100 % saturated limit cycle below ~6 m/s.** The wheel swings up to 46° peak-to-peak at ~0.5 Hz while creeping, turning at intersections and in parking lots. Turning NNLC off (nothing else changed) removes it: same car, same model, same day, low-speed chatter episodes go from **17.8/min to 3.0/min** and command saturation from **89–99 % to 1–2 %**.

Setup: comma four + chestnut (Radeon 9060), Toyota Prius Prime (`TOYOTA_PRIUS_TSS2`, torque control, NN model `TOYOTA_PRIUS_TSS2.json`), sunnypilot staging v2026.003.000. Lateral config as logged in `initData`: `NeuralNetworkLateralControl=1`, `EnforceTorqueControl=0` (→ `LatControlTorqueV0`), `LateralJerkTorqueController=0`, `LagdToggle=1` (learned lateralDelay 0.33 s), live torque params in use (latAccelFactor 1.42–1.56, friction 0.18). Big models: BMRLNAP v3/v4, Tee Time, Cinque Terre V2 (chestnut catalog); small model: WMI V12.

**Reproduce:** chestnut attached, any 2026 big model active, NNLC on, MADS engaged; creep at 1–5 m/s or take a slow turn. The torque command (`controlsState.lateralControlState.torqueState.output`) flips between +1.0 and −1.0 every ~1 s and the wheel oscillates. Set `NeuralNetworkLateralControl=0`, next ignition-on: gone.

#### 1. Baseline from 673 archived segments (Aug 27 – Sep 2, engaged and not steering-pressed; big = eGPU model, small = WMI V12 on the SoC; same roads)

| speed (m/s) | torque cmd saturated (\|cmd\| ≥ 0.99), big / small | high-passed wheel angle RMS (deg), big / small | model request swing (wheel-deg equiv.), big / small |
|---|---|---|---|
| 0–2 | 96 % / 89 % | 2.30 / 1.36 | 1.60 / 1.39 |
| 2–4 | 84 % / 45 % | 2.72 / 1.38 | 2.86 / 0.93 |
[...]

PR #2023, opened 2026-09-12
modeld: chestnut power limit, adjustable via sunnylink

PeterPhuTran:
**Description**

A chestnut powered from a 12V accessory outlet sits behind wiring built for 1–2 A dashcams. On my comma four (Prius Prime) the outlet path measures ~0.45 Ω (13.6 V no-load → 11.8 V at 3–4 A); at the GPU's stock boost transients (11 A+) the input collapses, which resets the USB link mid-transfer or hangs the device seconds into inference. The driving model itself draws 24–55 W.

This adds `ChestnutPowerLimit` (int, watts, default 0 = stock). When set, modeld sends `PPSMC_MSG_SetPptLimit` to the SMU right before the big-model transfer, reads the limit back and logs it. The applied value is already published as `chestnutState.powerLimitW`, so it shows up in telemetry. If the SMU call fails, the big-model load fails and modeld falls back to the small model like any other chestnut load error: an uncapped load on a supply the user has said cannot take it is the worse outcome.

Adjustable from sunnylink only (Device page, advanced controls, comma four), no on-device UI, following the CameraOffset precedent (#1614). Options Stock and 40–100 W (chestnut's 100 W inference design point); values are clamped to 40–100 W in code.

Files: param declaration, `modeld_v2/chestnut_power_limit.py` (mirrors the SMU access `ChestnutState.power_limit` already uses), a one-line hook in `load_big()`, the sunnylink device page (YAML + compiled JSON), unit tests.

**Verification**

comma four + chestnut (Radeon 9060 16 GB), Toyota Prius Prime, enclosure on the accessory outlet with a 12 AWG feed, running this cap as a fork patch on every drive since 2026-08-26:

- uncapped: `Device hang detected` ~46 s into engaged inference on every attempt, with USB resets during the 1.7 GB transfer
- 80 W: 27.5 min of continuous eGPU inference, zero model gaps, supply floor 11.7 V, draw peaks ~34 W
- 60 W: 88 min commute on the worst-case charging profile with zero hangs; 4/4 drives clean (~105 min); wall draw p50 40 W / max 59 W
- Cinque Terre V2 at 60 W: 36 ms median inference, 42 ms max, zero frame drops across 95 min of driving on 2026-09-11, so the cap costs no throughput
- `python sunnypilot/sunnylink/tools/compile_settings_ui.py --check` passes, ruff clean, new unit tests cover the clamp and the SMU message sequence against a stub device

Dongle ID and routes available on request; I keep them out of public threads. Same setup as #2019.

Assisted by Claude (disclosed in the commit per AI_POLICY.md). The cap has been in daily use on my device for three weeks.

### DIFF
--- a/openpilot/sunnypilot/sunnylink/settings_ui.json
+++ b/openpilot/sunnypilot/sunnylink/settings_ui.json
@@ -1742,7 +1742,7 @@
               "key": "ChestnutPowerLimit",
               "widget": "option",
               "title": "GPU Power Limit",
-              "description": "Caps the chestnut GPU's package power (PPT) before the driving model loads. Use it when the enclosure runs from a 12V accessory outlet, whose wiring sags under the GPU's stock boost transients and can reset the USB link or hang the model mid-drive. The driving model needs well under 60W. Stock leaves the GPU's own limit untouched. Takes effect at the next model load (ignition on).",
+              "description": "Caps the chestnut GPU's package power (PPT) before the driving model loads. Use it when the enclosure runs from a 12V accessory outlet, whose wiring sags under the GPU's stock boost transients and can reset the USB link or hang the model mid-drive. Current driving models run within 60W. Stock leaves the GPU's own limit untouched. Takes effect at the next model load (ignition on).",
               "options": [
                 {
                   "value": 0,
@@ -1771,14 +1771,6 @@
                 {
                   "value": 100,
                   "label": "100W"
-                },
-                {
-                  "value": 120,
-                  "label": "120W"
-                },
-                {
-                  "value": 150,
-                  "label": "150W"
                 }
               ],
               "visibility": [
--- a/openpilot/sunnypilot/sunnylink/settings_ui.json
+++ b/openpilot/sunnypilot/sunnylink/settings_ui.json
@@ -1734,6 +1734,71 @@
           ]
         },
         {
+          "id": "chestnut",
+          "title": "chestnut",
+          "description": "External GPU power",
+          "items": [
+            {
+              "key": "ChestnutPowerLimit",
+              "widget": "option",
+              "title": "GPU Power Limit",
+              "description": "Caps the chestnut GPU's package power (PPT) before the driving model loads. Use it when the enclosure runs from a 12V accessory outlet, whose wiring sags under the GPU's stock boost transients and can reset the USB link or hang the model mid-drive. The driving model needs well under 60W. Stock leaves the GPU's own limit untouched. Takes effect at the next model load (ignition on).",
+              "options": [
+                {
+                  "value": 0,
+                  "label": "Stock"
+                },
+                {
+                  "value": 40,
+                  "label": "40W"
+                },
+                {
+                  "value": 50,
+                  "label": "50W"
+                },
+                {
+                  "value": 60,
+                  "label": "60W"
+                },
+                {
+                  "value": 70,
+                  "label": "70W"
+                },
+                {
+                  "value": 80,
+                  "label": "80W"
+                },
+                {
+                  "value": 100,
+                  "label": "100W"
+                },
+                {
+                  "value": 120,
+                  "label": "120W"
+                },
+                {
+                  "value": 150,
+                  "label": "150W"
+                }
+              ],
+              "visibility": [
+                {
+                  "type": "capability",
+                  "field": "device_type",
+                  "equals": "mici"
+                }
+              ],
+              "enablement": [
+                {
+                  "type": "param",
+                  "key": "ShowAdvancedControls",
+                  "equals": true
+                }
+              ]
+            }
+          ]
+        },
+        {
           "id": "language",
           "title": "Language",
           "items": [
--- a/openpilot/sunnypilot/sunnylink/settings_ui_src/pages/device.yaml
+++ b/openpilot/sunnypilot/sunnylink/settings_ui_src/pages/device.yaml
@@ -68,7 +68,7 @@ sections:
     title: GPU Power Limit
     description: Caps the chestnut GPU's package power (PPT) before the driving model loads. Use it when the enclosure runs
       from a 12V accessory outlet, whose wiring sags under the GPU's stock boost transients and can reset the USB link or hang
-      the model mid-drive. The driving model needs well under 60W. Stock leaves the GPU's own limit untouched. Takes effect at
+      the model mid-drive. Current driving models run within 60W. Stock leaves the GPU's own limit untouched. Takes effect at
       the next model load (ignition on).
     options:
     - value: 0
@@ -85,10 +85,6 @@ sections:
       label: 80W
     - value: 100
       label: 100W
-    - value: 120
-      label: 120W
-    - value: 150
-      label: 150W
     visibility:
     - type: capability
       field: device_type
--- a/openpilot/sunnypilot/sunnylink/settings_ui_src/pages/device.yaml
+++ b/openpilot/sunnypilot/sunnylink/settings_ui_src/pages/device.yaml
@@ -59,6 +59,42 @@ sections:
       label: 24h
     - value: 1800
       label: 30h (Default)
+- id: chestnut
+  title: chestnut
+  description: External GPU power
+  items:
+  - key: ChestnutPowerLimit
+    widget: option
+    title: GPU Power Limit
+    description: Caps the chestnut GPU's package power (PPT) before the driving model loads. Use it when the enclosure runs
+      from a 12V accessory outlet, whose wiring sags under the GPU's stock boost transients and can reset the USB link or hang
+      the model mid-drive. The driving model needs well under 60W. Stock leaves the GPU's own limit untouched. Takes effect at
+      the next model load (ignition on).
+    options:
+    - value: 0
+      label: Stock
+    - value: 40
+      label: 40W
+    - value: 50
+      label: 50W
+    - value: 60
+      label: 60W
+    - value: 70
+      label: 70W
+    - value: 80
+      label: 80W
+    - value: 100
+      label: 100W
+    - value: 120
+      label: 120W
+    - value: 150
+      label: 150W
+    visibility:
+    - type: capability
+      field: device_type
+      equals: mici
+    enablement:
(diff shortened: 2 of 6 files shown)


openpilot/sunnypilot/sunnylink/settings_ui_src/pages/device.yaml
```diff
@@ -59,6 +59,42 @@ sections:
...
+    - value: 70
+      label: 70W
+    - value: 80
+      label: 80W
+    - value: 100
+      label: 100W
+    - value: 120
+      label: 120W
+    - value: 150
+      label: 150W
```
Discountchubbs: Chestnut is designed for 100W max inference. 120-150 isn't needed here.

openpilot/sunnypilot/sunnylink/settings_ui.json
```diff
@@ -1733,6 +1733,71 @@
...
+                  "value": 100,
+                  "label": "100W"
+                },
+                {
+                  "value": 120,
+                  "label": "120W"
+                },
+                {
+                  "value": 150,
+                  "label": "150W"
```
Discountchubbs: see above

Discountchubbs:
Please provide before and after comma connect routes for us to analyze

Discountchubbs (changes requested):
See comments and fix lint :)

PeterPhuTran:
Thanks for the review. All three items are addressed; details below so you have the full picture.

### 1. Options trimmed to chestnut's 100 W design point

- Pushed as d6dddadc9. sunnylink options are now **Stock, 40, 50, 60, 70, 80, 100 W**. The 120 W and 150 W entries are gone from `device.yaml` and the compiled `settings_ui.json`.
- The code clamp in `chestnut_power_limit.py` is now **40–100 W** (was 40–220), so a hand-edited param can't exceed 100 W either.
- For reference: the stock limit the SMU reports on my card (Radeon 9060 16 GB) is **160 W**, so every option in the list is a reduction.
- One question for you: given the 100 W design point, do you want the **default** to be 100 W instead of Stock (0)? I left it at 0 so the change is a no-op for existing chestnut installs and doesn't silently retune anyone's GPU on update. If you'd rather have every chestnut capped at its design point by default, that's a one-line change in `params_keys.h` and I'll make it.

### 2. Lint

`ty` was rejecting the test's stand-in `FakeParams` against the `Params | None` annotation on `get_power_limit()`. The tests now use a real `Params()` inside `OpenpilotTestCase`'s prefix and write the value through the typed API, so the param round-trip is exercised for real. `ty`, `ruff` and `compile_settings_ui.py --check` all pass locally; the unit-test job will confirm the rest.

While in there I tightened one sentence of the option description: "Current driving models run within 60W" replaces "needs well under 60W", because CTMV2 sits at 52–53 W under the cap (numbers below), so "well under" was no longer accurate.

### 3. Before / after routes

Constant across all of them: comma four + chestnut (Radeon 9060 16 GB), Toyota Prius Prime, enclosure fed from the switched 12 V accessory outlet. The outlet path measures ~0.45 Ω (13.6 V no-load → 11.8 V at 3–4 A).

I've DM'd you the dongle ID and the comma connect links for the 24 routes below, keyed by the labels in the tables (same policy as the PR body: identifiers stay out of the public thread). All of them are set to public in connect. `chestnutState` is in the qlog at 1 Hz (`supplyVoltage`, `supplyCurrent`, `powerDrawW`, `powerLimitW`, `gpuClockMhz`, `tempC`, `pcieLtssm`), so the enclosure input and the applied cap are viewable in cabana on every route. A hang shows up as the driving model dropping out mid-drive with the take-control alert; the tinygrad `Device hang detected` text only lives in the rlog / crash log, so say the word and I'll trigger rlog uploads for whichever of these you want.

**Before: no cap (SMU reports the stock 160 W)**

| Label | Date | Model | What happened |
|---|---|---|---|
| A1 | 2026-08-25 | BMV2 | Big-model load wedged mid-USB-transfer, with SuperSpeed resets in dmesg during the 1.7 GB copy. No model output for the 135 s session. |
| A2 (2 routes) | 2026-08-25 | BMV2 | Short evening sessions: the big model came up (one `chestnutState` sample: 13.4 V / 1.96 A / 24 W / limit 160 W), then sustained inference hung the GPU (`Device hang detected` + another SuperSpeed reset); the wedged modeld had to be SIGKILLed at car-off. |
| A3 | 2026-08-26 | BMV2 | The canonical failure. 27-min commute: BMV2 loaded in 38.5 s, I engaged, and after ~46 s of engaged GPU inference: `Device hang detected` → soft disable → fallback to the Qualcomm model for the rest of the drive. Good USB4 cable, so the cable is exonerated for this one. |

**After: 80 W cap**

| Label | Date | Model | What happened |
|---|---|---|---|
| B1 | 2026-08-26 | BMV2 | First clean big-model drive on the outlet. Cap requested=80 applied=80, 12 min engaged on and off, zero model gaps. Draw 24–30 W, GPU 2–7 %. |
| B2 | 2026-08-27 | BMV2 | 27.5 min of continuous eGPU inference, zero gaps. Supply 11.80–13.56 V, 2.3–4.2 A, 26–34 W across 148 `chestnutState` samples. |
| B3 | 2026-08-31 | BMV4 | **Hang at 80 W**, 17.6 min in: the car's DC-DC stepped 13.4 → 12.1 V, the enclosure input hit 11.57 V, `Device hang detected`. BMV4 was at 59 W of the 80 W cap, 65 °C, zero USB/PCIe errors. |
| B4 | 2026-08-31 | BMV4 | Same hang 2.3 min in, same charge→float step. |
| B5 | 2026-09-01 | BMV4 | Third hang, same mechanism. |

Why 80 W was not enough on this car: BMV4 is a heavier model than Tee Time (p95 64 W / max 73 W at the same cap, versus a flat 43–50 W), and all three hangs came 6–9 s after the DC-DC step-down, each preceded by the dataset's only ≥5 A enclosure bursts. Across 32 routes, 3 of 4 step-downs were fatal for BMV4 at 80 W versus 0 of 43 for everything else (Fisher p = 0.00025); Tee Time at 80 W survived 0 of 22. That is why I run 60 rather than 80, and why the option list starts at 40.

**After: 60 W cap (standing config since 2026-09-01, zero hangs since)**

| Label | Date | Model | What happened |
|---|---|---|---|
| C1 | 2026-09-01 | BMV4 | First drive at 60 W, rail below 12.5 V for 86 % of it: eGPU up on every segment. Wall max 55.3 W (−24 %), p95 50.5 W, clock −2 %. |
| C2 | 2026-09-01 | BMV4 | 88-min commute, rail below 12.5 V for 95 % of it including the same charging transition that killed B3: eGPU up 88/88 segments, zero hangs. p50 40.4 / p95 50.3 / max 58.9 W, peak 4.93 A. |
| C3 | 2026-09-03 | BMV4 | 39-min commute, first with the 12 AWG feed: 0 gaps, p50/p95/max 38.6/48.3/57.1 W, 4.72 A max. Capped record C1→C3: 0 hangs across ~22 charge→float step-downs. |
| C4 (5 routes) | 2026-09-07 | Cinque Terre | All clean, 0 hangs, load ~21 s, p50 ~40 W. The enclosure's own supply telemetry shows ~10 A / ~110 W bursts getting past the cap (caveat below); supply min ~10.4 V, still no hang. |
| C5 (3 routes) | 2026-09-08 | Cinque Terre | Same profile, all clean. |
| C6 (4 routes) | 2026-09-11 | CTMV2 | 6 starts, 100 % big-model frames, exec 36 ms p50 / 42 ms max, 0 frame drops over ~95 min. Enclosure-input draw p50 = p95 = 52–53 W (pinned at the cap, ~1650 MHz), peak supply current 9.8–10.2 A, ≤65 °C. |

One caveat that matters for your judgement: the SMU limit acts on its own averaged package-power estimate, so sub-second bursts still get through it (CTM/CTMV2 peak ~10 A at the enclosure input with `powerLimitW=60` confirmed in telemetry). It still removed the sustained ≥5 A tail that correlated with every hang, and the capped record since 2026-09-01 is zero hangs across every model I've run. On a bench supply none of this applies; the cap is for installs on marginal 12 V wiring, which is exactly the "12V car power cable" kit setup.

Happy to squash, rename, or move anything.

openpilot/sunnypilot/sunnylink/settings_ui_src/pages/device.yaml
PeterPhuTran: Done in d6dddadc9: options now end at 100W and the code clamp is 40–100 W. Stock on my 9060 16 GB reads back as 160 W, so every option is a reduction.

PeterPhuTran:
Update on logs: full rlogs are now in comma connect for the C4–C6 routes and one of the A2 routes, which is everything the device still had on disk. The remaining routes had already been rotated off the device, so their rlogs come from my archive instead: one tar per route in the standard `<route>--<segment>/rlog.zst` layout, linked in the DM alongside the route list. The README in that folder maps each tar to the labels above and points at the segments that hold each hang.

PeterPhuTran:
@Discountchubbs friendly re-review ping. Everything from your 09-12 review is in (options trimmed to Stock/40–100 W, clamp 40–100 W, `ty` fixed, tests on a real `Params`), master is merged in and CI is green. The routes are in the DM.

Ten more days of field data at the 60 W cap since the tables above, now on CTM:

- 16 drives, 475 segments, about 7.6 h of driving model output
- big model on 100 % of frames, no hang and no fallback to the small model on any drive
- 0.014 % of frames dropped, inference 35.9 ms median (worst per-segment median 37.5 ms)
- enclosure draw peaked at 54 W and 65 °C with `powerLimitW` 60 in every `chestnutState` sample

Another comma four + chestnut user landed on the same fix independently: character-s/openpilot@43847a3 caps the package at 80 W. On their setup the stock limit measured 170 W, and boost transients reached about 116 W at the enclosure input. That exceeds a 100 W PD charger, which drops out and takes the USB link with it ("Device hang detected"). So the failure isn't specific to my car's accessory-outlet wiring.

That leaves the question from my last comment: should the default stay Stock (a no-op on update) or be 100 W (the design point)? Either is a one-line change and I'm fine with both.

Assisted-by: Claude

PR #2026, opened 2026-09-13
models: don't freeze the ui on an unset LagdToggleDelay

AmyJeanes:
The reading of `LagdToggleDelay` with no parameter set freezes the UI as it has an incorrect second parameter which is fixed in this PR.

In practice this is unlikely to happen as the parameter is set during startup and only if live lag learning is disabled, but can bite if you're spinning up the UI directly without the manager backend running, which just bit me during testing.


Discountchubbs (approved):
Nice! This was due to the fact that before return_default was added, we could pass a default flag, now that return default is added, a default variable is done in param keys… nice job.

PR #2029, opened 2026-09-14
controls: add speed-dependent steering delay

datletrung:
## Summary

Addresses a steering-delay tuning issue where Live Steer Delay learning can become poorly suited to mixed driving conditions, particularly when the vehicle is driven primarily on the highway.

## Motivation

The Self Tune / Live Steer Delay behavior was observed to saturate around a delay that was well suited to sustained highway driving. Increasing the fixed delay to the maximum value improved low-speed steering behavior and eliminated a noticeable kick-out during turns.

However, using the maximum delay as a fixed value caused the vehicle to become too tight at highway speeds and introduced a tendency to ping-pong.

This highlights a limitation of using a single fixed steering delay across the full speed range:

* Lower delay can provide better highway behavior but less desirable low-speed turn response.
* Higher delay can improve low-speed behavior but make highway steering too tight.
* Live learning can converge toward a value that is more representative of the vehicle's predominant driving conditions rather than optimal across the full speed range.

This change addresses the issue by making the fixed steering delay speed-dependent.

## Changes

When Live Steer Delay is disabled:

* The configured steering delay is used as the base/highway delay.
* An additional configurable **City Delay Boost** is applied at lower speeds.
* City delay is capped at `0.50s`.
* The delay smoothly interpolates between the city and highway values from `50–80 km/h`.
* The existing Live Steer Delay learning path remains unchanged.

Default behavior:

* City Delay Boost: `0.30s`
* ≤50 km/h: up to `0.50s` total delay
* 50–80 km/h: smooth transition
* ≥80 km/h: configured base/highway delay

## User Configuration

Added **City Delay Boost** to:

* Device → Settings → Models
* SunnyLink → Models

The setting is configurable from `0.00–0.30s`.

## Validation

Vehicle: 2018 Toyota Camry

Testing with a fixed `0.50s` delay showed improved low-speed turn behavior and eliminated the observed kick-out.

However, the same fixed `0.50s` delay was too aggressive at highway speeds and resulted in ping-pong behavior.

The speed-dependent implementation is intended to preserve the improved low-speed behavior while returning to the configured lower delay at highway speeds.

Further before/after driving data and plots can be provided to validate steering response across the speed range.

## TODO

* Allow the user to configure the speed range over which the city/highway delay transition occurs by exposing the transition speed settings through the device and SunnyLink UI.


datletrung:
could be related to this issue upstream? https://github.com/commaai/openpilot/issues/34858

PR #2031, opened 2026-09-15
MADS: pause when brake is already held

jingshenSN2:
Similar to https://github.com/sunnypilot/sunnypilot/pull/1871.
When the brake is already held before MADS is enabled, once you enable MADS and keep holding brake for a while, there will be controlsMismatchLateral error and turn into immediate take control alert on the screen.
You can see the error from this route: b0d2bd701938150b/00000009--f20db3c785 (segment 1)

How to reproduce:
1. Set MADS Steering Mode on Brake to `Pause`
2. Start the car
3. Shift gear to D
4. Press and hold the brake pedal
5. Turn on Cruise Main while continuing to hold the brake
6. Wait about 2 seconds

Before the fix: a “Take Control Immediately” alert appears.
After the fix: MADS remains paused and no alert appears.


jingshenSN2:
Could a maintainer please remove the `ci` label? Mislabeled by the bot, and don't have permission to remove.

